Issue #32, opened 2020-02-17
Remove unneeded required 'app' parameter


Cy-4AH:
`app` parameter can be scanned by plugin themself from GoogleService configuration file of uploading binary

PR #33, opened 2020-02-18
Automatically detect google app id in iOS

Cy-4AH:
Resolves https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/32

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
@@ -1,4 +1,5 @@
 require 'fastlane_core/ui/ui'
+require 'cfpropertylist'
 
 module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?("UI")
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -56,6 +56,14 @@ def cleanup_tempfiles
         return if @tempfiles.nil?
         @tempfiles.each(&:unlink)
       end
+
+      def findout_ios_app_id_from_archive(path)
+        appPath = `plutil -extract ApplicationProperties xml1 -o - "#{path}/Info.plist" | plutil -convert json -o - - | jq -r ".ApplicationPath"`.delete("\n")
```
snatchev: fastlane has a dependency on the [plist gem](https://github.com/patsplat/plist). Please prefer that over shelling out to a binary if possible.

Cy-4AH: Plist gem doesn't parse binary plists. In archive folder GoogleServices-Info.plist is already compiled to binary plist.

snatchev: Ah, good catch! We also have the dependency on `CFPropertyList`

Cy-4AH: @snatchev, thanks, I didn't know about it. I have migrated on `CFPropertyList`

Cy-4AH:
@googlebot I signed it!

Cy-4AH:
@snatchev, hello! I doesn't have merge button.

snatchev:
@Cy-4AH thank you for the contribution (and ping)!

Issue #37, opened 2020-03-03
Configurable GoogleService-info.plist name


Robotex:
Regarding the PR #33, it would be nice if the filename could be configurable as in our company we have different names for each production stage

alexsinger:
Good idea! Feel free to submit a PR and we can take a look.

Robotex:
Just submitted a PR

Robotex:
Closing this issue as the PR has been merged in 0.2.2

PR #46, opened 2020-04-28
Allow passing custom path to GoogleService-Info.plist as parameter

Robotex:
Implements #37

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -26,7 +26,7 @@ module Fastlane
         elsif platform == :ios
           archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
           if archive_path
-            app_id = get_ios_app_id_from_archive(archive_path, params[:googleservice_info_plist_path])
+            app_id = get_ios_app_id_from_archive_plist(archive_path, params[:googleservice_info_plist_path])
           end
         end
         if app_id.nil?
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -26,7 +26,7 @@ module Fastlane
         elsif platform == :ios
           archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
           if archive_path
-            app_id = get_ios_app_id_from_archive(archive_path)
+            app_id = get_ios_app_id_from_archive(archive_path, params[:googleservice_info_plist_path])
           end
         end
         if app_id.nil?
@@ -111,6 +111,12 @@ module Fastlane
                                        verify_block: proc do |value|
                                          UI.user_error!("firebase_app_distribution: Couldn't find ipa file at path '#{value}'") unless File.exist?(value)
                                        end),
+          FastlaneCore::ConfigItem.new(key: :googleservice_info_plist_path,
+                                       env_name: "GOOGLESERVICE_INFO_PLIST_PATH",
+                                       description: "Path to your GoogleService-Info.plist file, relative to the root of your Xcode project",
+                                       default_value: "GoogleService-Info.plist",
+                                       optional: true,
+                                       type: String),
           # Android Specific
           FastlaneCore::ConfigItem.new(key: :apk_path,
                                        env_name: "FIREBASEAPPDISTRO_APK_PATH",
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
@@ -28,10 +28,10 @@ module Fastlane
         CFPropertyList.native_types(CFPropertyList::List.new(file: path).value)
       end
 
-      def get_ios_app_id_from_archive(path)
-        app_path = parse_plist("#{path}/Info.plist")["ApplicationProperties"]["ApplicationPath"]
-        UI.shell_error!("can't extract application path from Info.plist at #{path}") if app_path.empty?
-        identifier = parse_plist("#{path}/Products/#{app_path}/GoogleService-Info.plist")["GOOGLE_APP_ID"]
+      def get_ios_app_id_from_archive_plist(archive_path, plist_path)
+        app_path = parse_plist("#{archive_path}/Info.plist")["ApplicationProperties"]["ApplicationPath"]
+        UI.shell_error!("can't extract application path from Info.plist at #{archive_path}") if app_path.empty?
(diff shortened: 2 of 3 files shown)


Robotex:
@googlebot I signed it!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -84,6 +84,12 @@ def self.available_options
                                        verify_block: proc do |value|
                                          UI.user_error!("firebase_app_distribution: Couldn't find ipa file at path '#{value}'") unless File.exist?(value)
                                        end),
+          FastlaneCore::ConfigItem.new(key: :gsp_path,
+                                       env_name: "GOOGLESERVICE_PLIST_PATH",
```
jladieu, line 88: Consider GOOGLESERVICE_INFO_PLIST_PATH, based on my (very brief googling) I think this file is more commonly referred to as GoogleService-Info (_with_  the info).

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -84,6 +84,12 @@ def self.available_options
                                        verify_block: proc do |value|
                                          UI.user_error!("firebase_app_distribution: Couldn't find ipa file at path '#{value}'") unless File.exist?(value)
                                        end),
+          FastlaneCore::ConfigItem.new(key: :gsp_path,
```
jladieu, line 87: Rather than abbreviate GSP, consider expanding this out to be called `googleservice_info_plist_path`; this name is a bit more intention revealing and doesn't require memorizing a new acronym.  It will also hopefully make it easier to grep for those doing exploratory review and maintain consistency with the environment variable naming I suggested.

Robotex: > Rather than abbreviate GSP, consider expanding this out to be called `googleservice_info_plist_path`; this name is a bit more intention revealing and doesn't require memorizing a new acronym. It will also hopefully make it easier to grep for those doing exploratory review and maintain consistency with the environment variable naming I suggested.

I initially picked `gsp_path` because it was already used in upload_symbols_to_crashlytics but I will make the suggested changes.

shinriyo:
Conflicting files

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -84,6 +84,12 @@ def self.available_options
                                        verify_block: proc do |value|
                                          UI.user_error!("firebase_app_distribution: Couldn't find ipa file at path '#{value}'") unless File.exist?(value)
                                        end),
+          FastlaneCore::ConfigItem.new(key: :googleservice_info_plist_path,
```
lfkellogg, line 87: Is this parameter generally going to be used to change only the name of the file? Or will it be used to set the path relative to the project root?

If it's the former, can you please change "path" to "name" throughout the ConfigItem declaration and its uses?

If it's the latter, can you update the description to be "Path to your GoogleService-Info.plist file, relative to the root of your Xcode project"

lfkellogg (changes requested):
Hi @Robotex,

Sorry for the slow back and forth on this. As you may have noticed we recently merged an overhaul of the plugin that removes its dependency on the Firebase CLI, hence the new merge conflicts on this PR.

Luckily it should be pretty easy to resolve those. The method you changed has simply been renamed and is now called here: https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/master/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb#L33.

Could you resolve the conflicts and also take a look at the comment I left?

Thanks,
Lee

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
Robotex: I updated the description to _"Path to your GoogleService-Info.plist file, relative to the root of your Xcode project"_, because I find it to describe a more flexible usage IMHO, even though in the company I work at we use it to change the name of the file only

lfkellogg: Sounds good!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -26,7 +26,7 @@ def self.run(params)
         elsif platform == :ios
           archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
           if archive_path
-            app_id = get_ios_app_id_from_archive(archive_path)
+            app_id = get_ios_app_id_from_archive(archive_path, params[:googleservice_info_plist_path])
```
lfkellogg, line 29: You need to update this call to use the updated method name.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -28,10 +28,10 @@ def parse_plist(path)
         CFPropertyList.native_types(CFPropertyList::List.new(file: path).value)
       end
 
-      def get_ios_app_id_from_archive(path)
-        app_path = parse_plist("#{path}/Info.plist")["ApplicationProperties"]["ApplicationPath"]
-        UI.shell_error!("can't extract application path from Info.plist at #{path}") if app_path.empty?
-        identifier = parse_plist("#{path}/Products/#{app_path}/GoogleService-Info.plist")["GOOGLE_APP_ID"]
+      def get_ios_app_id_from_archive_plist(archive_path, plist_path)
```
lfkellogg, line 31: I'm not sure why the tests didn't run against this PR (maybe because it's from before we had CI hooked up) but could you please update the spec here to include the new `plist_path` argument? https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/master/spec/firebase_app_distribution_helper_spec.rb#L75

You can run the specs by running `rake` from the plugin project root.

lfkellogg (changes requested):

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
Robotex: I updated the spec and now launching `bundle exec rake -t` passes the tests, however I have no experience in RSpec so please correct me if I'm mistaken

lfkellogg: Looks great, thanks!

lfkellogg (approved):
Thanks again!

As I'm sure you noticed when testing, the bug you posted is preventing this from running without the app ID param set: https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/154

That will be fixed with https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/pull/155 (and I've verified locally that it does) so I'll go ahead and merge this and rebase the other PR.

shinriyo:
Thank you!

PR #53, opened 2020-06-03
Login

mannyjimenez0810:
Added in logging using Google Auth and setting the FIREBASE_TOKEN environmental variable.


lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
+require 'googleauth'
+require 'googleauth/stores/file_token_store'
+require "google/apis/people_v1"
+# require "googleauth"
+# require "googleauth/stores/file_token_store"
```
lfkellogg, line 5: Can you please remove these commented-out imports?

fastlane/Fastfile
```diff
@@ -1,3 +0,0 @@
-lane :test do
-  firebase_app_distribution
-end
```
lfkellogg, line 3: I think I remember when this and fastlane/Pluginfile were deleted, but I don't remember why. Is there a particular reason? If not I think we should restore them. Looks like it has something to do with running tests.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+require "fileutils"
+module Fastlane
```
lfkellogg, line 7: I think having an empty line between the `require` statements and the `module` would create some visual separation between them.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+  module Actions
+    module SharedValues
+      TEST_AUTH_CUSTOM_VALUE = :TEST_AUTH_CUSTOM_VALUE
```
lfkellogg, line 10: It looks like this is unused. Can we remove this variable and/or the entire SharedValues module?

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    class LoginAction < Action
+      def self.run(params)
+        client_id = Google::Auth::ClientId.new("563584335869-fgrhgmd47bqnekij5i8b5pr03ho849e6.apps.googleusercontent.com", "j9iVZfS8kkCEFUPaAeJV0sAi")
```
lfkellogg, line 15: I think it would be good for readability if these two strings were put into constant class variables: https://www.rubyguides.com/2017/07/ruby-constants/

It would help the reader identify what the two strings are, and would improve the readability of this line.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+        scope = 'https://www.googleapis.com/auth/cloud-platform'
```
lfkellogg, line 16: This could also be a class-level constant, and then you wouldn't even need the scope variable.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+        token_store = nil
+        authorizer = Google::Auth::UserAuthorizer.new(client_id, scope, token_store)
+        url = authorizer.get_authorization_url(base_url: "urn:ietf:wg:oauth:2.0:oob")
```
lfkellogg, line 19: This base url would especially be nice as a class-level constant, since it's used twice.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        UI.message("Please open the following address in your browser:")
+        UI.message(url.to_s)
+        UI.message(" ")
```
lfkellogg, line 23: I think an empty string `""` would be better here unless you need the space.

By the way it looks like putting empty UI.message('') calls like this is used within the plugin, which reinforces to me that this was a good approach :+1:: https://github.com/fastlane/fastlane/blob/80a59306fb32b3c29aa5a2e548045c6eb5227f73/fastlane/lib/fastlane/setup/setup.rb#L42

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+        code = UI.input("Enter the resulting code here: ")
+        credentials = authorizer.get_credentials_from_code(code: code, base_url: "urn:ietf:wg:oauth:2.0:oob")
+        token = credentials.refresh_token
+
+        UI.message("Refresh Token: #{token}")
+        UI.message(" ")
+        UI.message("Set the refresh token as a FIREBASE_TOKEN enviorment variable")
```
lfkellogg, line 30: Typo: `environment`

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+      rescue
```
lfkellogg, line 31: I think it would be good to have two `rescue` clauses here. One that rescues just the specific authentication error that you got when entering an invalid or malformed code. The other would rescue any other error.

The message for the first could be something like:  "The code you entered was invalid. Ensure that you have copied the code correctly."

The message for the second could be more general, i.e.: "An error occurred."

Also, for these messages you should use UI.error, and consider using UI.crash for the second case, since some sort of unexpected error happened in that case. See [this guide]( https://docs.fastlane.tools/advanced/actions/#user-input-and-output) and see [this example](https://github.com/fastlane/fastlane/blob/334d3789c37ad0af236fdd39696d7684bf41bba7/match/lib/match/encryption/openssl.rb#L131) as well.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -59,13 +59,14 @@ def cleanup_tempfiles
       end
 
       def parse_plist(path)
-        CFPropertyList.native_types(CFPropertyList::List.new(:file => path).value)
+        CFPropertyList.native_types(CFPropertyList::List.new(file: path).value)
```
lfkellogg, line 62: Interesting changes in this file: did they get changed by the linter or did you make these?

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+        UI.message("An error has occured please login again. Ensure that you have copied the code correctly")
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Allow the user to login by getting a refresh token from the web and then setting an environmental variable to that token "
```
lfkellogg, line 40: I think this should start with the active verb, like "Log in by getting..." (I think log in vs. login?)

Also please run this copy by Alex, since this will be user facing and so should be approved by the product manager. I'd recommend either sending him a link to this PR or send it to him via chat.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      def self.details
+        "You can use this action to do not have to login multiple times."
```
lfkellogg, line 44: The details should also be confirmed with Alex. I'm not sure what normally goes in the details for an action, but this could also potentially include a link to the App Distribution docs.

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      def self.available_options
```
lfkellogg, line 47: Can you see if you can just remove these empty method definitions?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
mannyjimenez0810: Yea that must have been linter since  I did not change helper at all.

Gemfile
```diff
@@ -1,6 +1,7 @@
 source('https://rubygems.org')
 
 gemspec
+gem 'google-api-client', '~> 0.34'
```
tagboola, line 4: Any reason for not using the latest version `0.4.0`?

lib/fastlane/plugin/firebase_app_distribution/actions/login.rb
```diff
@@ -0,0 +1,65 @@
...
+# require "googleauth"
+# require "googleauth/stores/file_token_store"
+require "fileutils"
+module Fastlane
+  module Actions
+    module SharedValues
+      TEST_AUTH_CUSTOM_VALUE = :TEST_AUTH_CUSTOM_VALUE
+    end
+
+    class LoginAction < Action
```
tagboola, line 13: The name `login` is probably too generic of a name for the action. We should probably make this class `FirebaseAppDistributionLogin` so the action is named `firebase_app_distribution_login`. Or maybe there is a way to set an action name that's different from the class?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,57 @@
+require 'googleauth'
+require 'googleauth/stores/file_token_store'
+require "google/apis/people_v1"
+require "fileutils"
+
+module Fastlane
+  module Actions
+    class LoginAction < Action
+      OOB_URI = "urn:ietf:wg:oauth:2.0:oob"
+      SCOPE = 'https://www.googleapis.com/auth/cloud-platform'
```
lfkellogg, line 10: Can you use double-quotes everywhere?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,57 @@
...
+      CLIENT_ID = "563584335869-fgrhgmd47bqnekij5i8b5pr03ho849e6.apps.googleusercontent.com"
+      CLIENT_SECRET = "j9iVZfS8kkCEFUPaAeJV0sAi"
+
+      def self.run(params)
+        client_id = Google::Auth::ClientId.new(CLIENT_ID,CLIENT_SECRET)
```
lfkellogg, line 15: Please put a space between these args

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,57 @@
...
+        token_store = nil
+        authorizer = Google::Auth::UserAuthorizer.new(client_id, SCOPE, token_store)
+        url = authorizer.get_authorization_url(base_url: "urn:ietf:wg:oauth:2.0:oob")
```
lfkellogg, line 18: Use your class variable here

lfkellogg:
By the way, even though I left comments, I went ahead and approved because I consider these comments pretty small and trust you to revise them :)

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      def self.details
+        "You can use this action to not have to login multiple times."
+      end
+
+      def self.authors
+        ["Manny Jimenez Github: mannyjimenez0810, Alonso Salas Infante Github: alonsosalasinfante"]
+      end
+
+      def self.is_supported?(platform)
```
lfkellogg, line 53: Based on Tunde's suggestion I think we should set this to only support ios and Android, using the syntax described in the comment here: https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/bb0dfea89fb4cb62ce8e63722ae98afc4234280c/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb#L163

While you're at it you might want to change the upload action to only support those platforms as well.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,57 @@
...
+        UI.error(error.to_s)
+        UI.crash!("An error has occured please login again.")
+      end
+
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Login by getting a refresh token from the web and then setting an environmental variable to that token."
```
alexsinger, line 42: What about: "Authenticate with Firebase App Distribution using a Google account" (is <--- accurate)?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      def self.details
+        "You can use this action to not have to login multiple times."
```
alexsinger, line 46:
```
"Log in to Firebase App Distribution using a Google account to generate an authentication token. This token is associated with an environment variable and used to authenticate with your Firebase project.", 
"See <some link we'll provide> for more information."
```

alexsinger: Thoughts? ^

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
mannyjimenez0810: I believe that is accurate will change now.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
mannyjimenez0810: I believe saying the token is stored in an environmental variable instead of associated might be clearer on how it works.

alexsinger: sgtm

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,59 @@
+require 'googleauth'
+require 'googleauth/stores/file_token_store'
+require "google/apis/people_v1"
+require "fileutils"
+
+module Fastlane
+  module Actions
+    class FirebaseAppDistributionLoginAction < Action
+      OOB_URI = "urn:ietf:wg:oauth:2.0:oob"
+      SCOPE = 'https://www.googleapis.com/auth/cloud-platform'
```
tagboola, line 10: This should be in double quotes as well to be consistent with the other constants.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,59 @@
...
+      CLIENT_ID = "563584335869-fgrhgmd47bqnekij5i8b5pr03ho849e6.apps.googleusercontent.com"
+      CLIENT_SECRET = "j9iVZfS8kkCEFUPaAeJV0sAi"
+
+      def self.run(params)
+        client_id = Google::Auth::ClientId.new(CLIENT_ID,CLIENT_SECRET)
+        token_store = nil
```
tagboola, line 16: Because this variable isn't used anywhere else consider just passing `nil` to the `Google::Auth::UserAuthorizer.new` method.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,59 @@
...
+        authorizer = Google::Auth::UserAuthorizer.new(client_id, SCOPE, token_store)
+        url = authorizer.get_authorization_url(base_url: OOB_URI)
+
+        UI.message("Please open the following address in your browser:")
+        UI.message(url.to_s)
```
tagboola, line 21: The `get_authorization_url` already converts the url to a string so the `to_s` is unnecessary:
https://github.com/googleapis/google-auth-library-ruby/blob/9bcac01574d539e281326beaa25d19e07002b525/lib/googleauth/user_authorizer.rb#L114

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,59 @@
...
+        UI.message("")
+        code = UI.input("Enter the resulting code here: ")
+        credentials = authorizer.get_credentials_from_code(code: code, base_url: OOB_URI)
+        token = credentials.refresh_token
+
+        UI.message("Refresh Token: #{token}")
```
tagboola, line 27: The `token` variable isn't used anywhere else so consider inlining i.e. `UI.message("Refresh Token: #{credentials.refresh_token}")`.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,59 @@
...
+      #####################################################
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Authenticate with Firebase App Distribution using a Google account."
+      end
+
+      def self.details
+        ["Log in to Firebase App Distribution using a Google account to generate an authentication token. This token is stored within an environment variable and used to authenticate with your Firebase project.",
```
tagboola, line 46: I _think_ you want to return a string here, not an array i.e.

```
"Log in to Firebase App Distribution using a Google account to generate an authentication token. This token is stored within an environment variable and used to authenticate with your Firebase project. See https://firebase.google.com/docs/app-distribution/ios/distribute-fastlane for more information."
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -0,0 +1,58 @@
...
+      # @!group Documentation
+      #####################################################
+
+      def self.description
+        "Authenticate with Firebase App Distribution using a Google account."
+      end
+
+      def self.details
+        "Log in to Firebase App Distribution using a Google account to generate an authentication "\
+        "token. This token is stored within an environment variable and used to authenticate with your Firebase project."\
```
tagboola, line 45: Need at the end of this line i.e.  `"Firebase project. "\`

PR #56, opened 2020-06-08
Getapp Portion of Upload Flow

alonsosalasinfante:
Added GET app portion of upload flow with some error handling that will be subject to change


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +165,52 @@ def self.is_firebasecmd_supported?(cmd)
...
+          conn.response(:raise_error)
+          conn.adapter(Faraday.default_adapter)
+        end
+
+        response = connection.get(url) do |request|
+          request.headers["Authorization"] = "Bearer " + token
+        end
+
+        contactEmail = response.body[:contactEmail]
+        UI.message(contactEmail)
```
lfkellogg, line 202: Remove this

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +165,52 @@ def self.is_firebasecmd_supported?(cmd)
...
+
+        if contactEmail.strip.empty?
+          UI.error("Empty contact email")
+        end
+      rescue => error
+        if error.class == Faraday::ResourceNotFound
```
lfkellogg, line 208: Instead of this `if`, you should rescue a specific error class: https://blog.appsignal.com/2018/04/10/rescuing-exceptions-in-ruby.html

You can still have a second, generic `rescue` after that one to catch any other errors.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +165,52 @@ def self.is_firebasecmd_supported?(cmd)
...
+
+        response = connection.get(url) do |request|
+          request.headers["Authorization"] = "Bearer " + token
+        end
+
+        contactEmail = response.body[:contactEmail]
+        UI.message(contactEmail)
+
+        if contactEmail.strip.empty?
+          UI.error("Empty contact email")
```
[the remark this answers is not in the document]
lfkellogg: Thinking ahead to how this method will be used: should there be a return value from this method? What would it be in this case, where the app is invalid because there's no contact email? Or maybe the return value of this method is the app itself, and this check happens in the main body of the action.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +165,52 @@ def self.is_firebasecmd_supported?(cmd)
...
+        end
+      rescue => error
+        if error.class == Faraday::ResourceNotFound
+          UI.crash!("Failed to onboard.")
+        else
+          UI.crash!("Failed to fetch app information")
```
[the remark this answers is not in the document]
lfkellogg: Looks like it should always have a message: https://ruby-doc.org/core-2.2.0/Exception.html#method-i-message

fastlane/Fastfile
```diff
@@ -1,3 +1,9 @@
 lane :test do
   firebase_app_distribution
 end
+
```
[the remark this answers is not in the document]
alonsosalasinfante: Should I remove the entire Fastfile?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -169,47 +170,42 @@ def self.is_firebasecmd_supported?(cmd)
...
 
         contactEmail = response.body[:contactEmail]
-        UI.message(contactEmail)
 
         if contactEmail.strip.empty?
-          UI.error("Empty contact email")
+          UI.error("We could not find a contact email for app #{appId}. Please visit App Distribution within the Firebase Console to set one up.")
         end
+      rescue Faraday::ResourceNotFound => error
+        UI.crash!("App Distribution could not find your app #{params[:app]}. Make sure to onboard your app by pressing the "Get started" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
```
lfkellogg: Looks like the double-quotes around `Get started` need to be escaped: `"something like \"this\" I believe"`

lfkellogg: Also, I believe this link will take them to the last project the loaded in the Firebase console. When we get the app back, I assume we get a project ID back as well (not the project number, but the ID), right? Could we link to that specific project? I think you could just replace the `_` with the project ID.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,47 @@ def self.is_firebasecmd_supported?(cmd)
...
+        return client.access_token
+      rescue Signet::AuthorizationError => error
+        UI.crash!("Wrong value for FIREBASE_TOKEN")
+      end
+
+      def self.get_app(params)
+        token = get_token
+
+        # begin
+        appId = (params[:app]).to_s
```
tagboola, line 188: Rename to `app_id`.

Also, no need to call `to_s` here. `app_id` is only used in interpolated strings, and ruby automatically calls `to_s` on values used in string interpolation.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,47 @@ def self.is_firebasecmd_supported?(cmd)
...
+          conn.response(:json, parser_options: { symbolize_names: true })
+          conn.response(:raise_error) # raise_error middleware will run before the json middleware
+          conn.adapter(Faraday.default_adapter)
+        end
+
+        response = connection.get("#{PATH}#{appId}") do |request|
+          request.headers["Authorization"] = "Bearer " + token
+        end
+
+        contactEmail = response.body[:contactEmail]
```
tagboola, line 200: `contact_email`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,47 @@ def self.is_firebasecmd_supported?(cmd)
...
+        client.fetch_access_token!
+        return client.access_token
+      rescue Signet::AuthorizationError => error
+        UI.crash!("Wrong value for FIREBASE_TOKEN")
+      end
+
+      def self.get_app(params)
+        token = get_token
+
+        # begin
```
tagboola, line 187: Remove unused comment.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,47 @@ def self.is_firebasecmd_supported?(cmd)
...
+        appId = (params[:app]).to_s
+
+        connection = Faraday.new(url: BASE_URL) do |conn|
+          conn.response(:json, parser_options: { symbolize_names: true })
+          conn.response(:raise_error) # raise_error middleware will run before the json middleware
+          conn.adapter(Faraday.default_adapter)
+        end
+
+        response = connection.get("#{PATH}#{appId}") do |request|
```
tagboola, line 196: We only want to catch errors thrown by the `GET` request, so we should surround it in a begin clause and only rescue from errors thrown by the `connection.get` method i.e.

```
begin
  response = connection.get("#{PATH}#{appId}") do |request|
  request.headers["Authorization"] = "Bearer " + token
rescue Faraday::ResourceNotFound => error
  UI.user_error!("App Distribution could not find your app #{params[:app]}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
rescue => error
  UI.crash!("Failed to fetch app information: #{error.message}")
end

if contact_email.strip.empty?
  UI.user_error("We could not find a contact email for app #{appId}. Please visit App Distribution within the Firebase Console to set one up.")
end
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,47 @@ def self.is_firebasecmd_supported?(cmd)
...
+          request.headers["Authorization"] = "Bearer " + token
+        end
+
+        contactEmail = response.body[:contactEmail]
+
+        if contactEmail.strip.empty?
+          UI.error("We could not find a contact email for app #{appId}. Please visit App Distribution within the Firebase Console to set one up.")
```
tagboola, line 203: I've updated in the errors used in the comment above, but here is some guidance on what errors to use:
https://docs.fastlane.tools/advanced/actions/#user-input-and-output

Based on that, I think an empty email and resource not found errors should used `UI.user_error!` and any other network error should use `UI.crash!`.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,47 @@ def self.is_firebasecmd_supported?(cmd)
...
+          refresh_token: ENV["FIREBASE_TOKEN"]
+        )
+
+        client.fetch_access_token!
+        return client.access_token
+      rescue Signet::AuthorizationError => error
+        UI.crash!("Wrong value for FIREBASE_TOKEN")
+      end
+
+      def self.get_app(params)
```
tagboola, line 184: Consider renaming to `validate_app!` since we aren't actually returning a value but just validating the app exists and has a contact email.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -181,31 +181,30 @@ def self.get_token
         UI.crash!("Wrong value for FIREBASE_TOKEN")
       end
 
-      def self.get_app(params)
+      def self.validate_app!(params)
```
tagboola, line 184: nit: Consider just passing the `app_id` since it's the only value in the `params` hash that is used.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -191,6 +166,46 @@ def self.is_firebasecmd_supported?(cmd)
...
+          conn.response(:json, parser_options: { symbolize_names: true })
+          conn.response(:raise_error) # raise_error middleware will run before the json middleware
+          conn.adapter(Faraday.default_adapter)
+        end
+        begin
+          response = connection.get("#{PATH}#{app_id}") do |request|
+            request.headers["Authorization"] = "Bearer " + token
+          end
+        rescue Faraday::ResourceNotFound => error
+          UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/#{app_id}/appdistribution")
```
lfkellogg, line 198: This uses the App ID in the URL, when what you really need is the Project ID (Projects can contain many Apps). I now realize that in this case you don't actually have the Project ID, since you were unable to fetch the app information. So please disregard my original comment, sorry about that. `https://console.firebase.google.com/project/_/appdistribution` is fine.

PR #57, opened 2020-06-18
Upload status

alonsosalasinfante:
Add upload binary and upload status polling


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,12 +15,40 @@ class FirebaseAppDistributionAction < Action
       FIREBASECMD_ACTION = "appdistribution:distribute".freeze
       BASE_URL = "https://firebaseappdistribution.googleapis.com"
       PATH = "/v1alpha/apps/"
+      MAX_POLLING_RETRIES = 60
+      POLLING_INTERVAL_S = 2
+      CONNECTION = Faraday.new(url: BASE_URL) do |conn|
```
lfkellogg, line 20: I think your instinct to share this connection is good, and looks like it should work. However, using an all-caps variable name here makes this a "constant" variable, which is shared between all instances of the class. That seems like overkill, and I think it's better to keep one connection per instance of the class, to isolate the behavior. In order to do that you can use an "instance" variable, which starts with a "@" like `@connection` for example.

In this case it might be good to use a pattern I've seen folks use a lot:
```
def connection
  @connection ||= Faraday.new(...
end
```
And then when you want to use a connection you just use `connection`. This calls the method, which will either return the `@connection` instance variable if it exists, or will create it if it does not, and then return it. This has the added benefit of making it "lazy" as well, meaning it doesn't actually create the connection object until you use it. See more [here](http://blog.jayfields.com/2007/07/ruby-lazily-initialized-attributes.html).

Let me know if you want to chat this out.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,12 +15,40 @@ class FirebaseAppDistributionAction < Action
...
+        conn.response(:raise_error) # raise_error middleware will run before the json middleware
+        conn.adapter(Faraday.default_adapter)
+      end
 
       extend Helper::FirebaseAppDistributionHelper
 
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
-        validate_app!(params[:app])
+        token_string = validate_app!(params[:app])
```
lfkellogg, line 30: I think you need to consider what happens when there is an error in `validate_app!`, either if the request fails or there's no email.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,12 +15,40 @@ class FirebaseAppDistributionAction < Action
...
+        begin
+          for checks in 1..MAX_POLLING_RETRIES
```
lfkellogg, line 32: Check out "The Times Loop" here: https://www.rubyguides.com/ruby-tutorial/loops/

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,12 +15,40 @@ class FirebaseAppDistributionAction < Action
...
+            status = upload_status!(token_string, params[:app], params[:apk_path])
+            if status == "SUCCESS"
+              UI.message("App Uploaded Successfully!") # Probably want something like it had before of apk has already been uploaded so no need to upload again
```
lfkellogg, line 35: Can this just match either the messaging in the Gradle plugin or the Firebase CLI?

lfkellogg: Oh I see, do you want to return one of two different messages, depending on whether the app has been uploaded or not during this execution?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
         contact_email = response.body[:contactEmail]
 
         if contact_email.strip.empty?
           UI.user_error!("We could not find a contact email for app #{app_id}. Please visit App Distribution within the Firebase Console to set one up.")
         end
+        return "projects/#{response.body[:projectNumber]}/apps/#{response.body[:appId]}/releases/-/binaries/"
+      end
+
+      def self.upload_binary!(app_id, binary_path)
+        token = get_token
```
lfkellogg, line 232: Since you're using get_token in a number of places I think the "lazy" approach I mentioned above would be good for this method as well, that way you're not making a fresh request (and fetching a new access token) every time.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
+        begin
+          response = CONNECTION.post("/app-binary-uploads?app_id=#{app_id}", File.open(binary_path).read) do |request|
+            request.headers["Authorization"] = "Bearer " + token
+            # request.headers["X-APP-DISTRO-API-CLIENT-ID"] = "FastLane"
```
lfkellogg, line 236: Interesting, is this for analytics?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
+          end
+          response.status
```
lfkellogg, line 238: Do you need to return this status?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
+        rescue Faraday::ResourceNotFound => error
+          UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
```
lfkellogg, line 240: You use this in another place right? Since this is so long and complex, having it in one shared place could be useful.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
+        rescue => error
+          UI.crash!("Failed to upload distribution: #{error.message}")
+        end
+      end
+
+      def self.upload_status!(app_token, app_id, binary_path)
+        token = get_token
+        upload_token = CGI.escape(app_token + Digest::SHA256.hexdigest(File.open(binary_path).read))
```
lfkellogg, line 248: Since you call `upload_status!` once per iteration while polling, it would be good to avoid re-reading and hashing the entire binary each time through the loop.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
+
+        begin
+          response = CONNECTION.get("#{PATH}#{app_id}/upload_status/#{upload_token}") do |request|
+            request.headers["Authorization"] = "Bearer " + token
+          end
+          response.body[:status]
+        rescue => error
+          UI.crash!(" #{error.message}")
```
lfkellogg, line 256: Do you want to add some context to the message here?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: +1 Great description!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: Since this return the upload token maybe this should be called `get_upload_token`. Also my bad I think mis-lead you with the "bang" (`!`)character in `validate_app!`. The "bang" should actually be used when the method modifies the object so `get_upload_token` doesn't need it :slight

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: +1

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: +1. Remove the "bang" (`!`) based on my previous comment.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: Yep! Is the plan to add the analytic headers in a separate PR? If so you can remove this line 🙂

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -183,27 +211,50 @@ def self.get_token
...
+          end
+          response.status
+        rescue Faraday::ResourceNotFound => error
+          UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
+        rescue => error
+          UI.crash!("Failed to upload distribution: #{error.message}")
+        end
+      end
+
+      def self.upload_status!(app_token, app_id, binary_path)
```
tagboola, line 246: Remove the "bang" (`!`) based on my previous comment.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: Yes this is for analytics. I know Tunde said we would have the Fastlane client id available, but we just kept it commented until we got confirmation on what values to set the metrics to.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,12 +15,40 @@ class FirebaseAppDistributionAction < Action
...
-        validate_app!(params[:app])
+        token_string = validate_app!(params[:app])
+        begin
+          for checks in 1..MAX_POLLING_RETRIES
+            status = upload_status!(token_string, params[:app], params[:apk_path])
+            if status == "SUCCESS"
+              UI.message("App Uploaded Successfully!") # Probably want something like it had before of apk has already been uploaded so no need to upload again
+              break
+            elsif status == "IN_PROGRESS"
+              UI.message("Still in progress...")
```
tagboola, line 38: Will this print out after each poll? I think we only need to print this out a status once before we upload i.e `Uploading APK to Firebase App Distribution...`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: +1

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,71 @@ class FirebaseAppDistributionAction < Action
...
+      POLLING_INTERVAL_S = 2
 
       extend Helper::FirebaseAppDistributionHelper
 
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
-        validate_app!(params[:app])
+        app_id = nil
+        platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
+        binary_path = Shellwords.escape(params[:ipa_path] || params[:apk_path])
```
tagboola, line 28: Remove `Shellwords.escape`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,71 @@ class FirebaseAppDistributionAction < Action
...
+          app_id = params[:app]
+        else
+          if platform == :ios
+            archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
+            if archive_patharchive_path
+              app_id = findout_ios_app_id_from_archive(archive_path)
+            end
+          end
+          if app_id.nil?
+            app_id = UI.input("Your app's Firebase App ID. You can find the App ID in the Firebase console, on the General Settings page: ")
```
tagboola, line 40: Do we need to do this? Based on the comment above, my understanding was that the call to `params.values` checked the parameter values and if the app id is blank, prompted the user with the same message.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,71 @@ class FirebaseAppDistributionAction < Action
...
+          end
+        end
+
+        upload_token = get_upload_token(app_id, binary_path)
```
tagboola, line 44: Before we polling we should make an initial `upload_status` call outside of the loop. If that returns `SUCCESS` then we should display "This APK has been uploaded before. Skipping upload step." This is what we do in the gradle plugin (see attached screenshot).
Screen Shot 2020-06-22 at 3 32 13 PM

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,71 @@ class FirebaseAppDistributionAction < Action
...
+        MAX_POLLING_RETRIES.times do
+          status = upload_status(upload_token, app_id)
+          if status == "SUCCESS"
+            UI.message("App Uploaded Successfully!")
+            break
+          elsif status == "IN_PROGRESS"
+            sleep(POLLING_INTERVAL_S)
+          else
+            UI.message("Uploading APK")
```
tagboola, line 53: If the binary has not been uploaded before, let's show the same two messages that the gradle plugin shows i.e.
```
UI.message("This APK has not been uploaded before.")
UI.message("Uploading the APK.")
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,71 @@ class FirebaseAppDistributionAction < Action
...
+          if app_id.nil?
+            app_id = UI.input("Your app's Firebase App ID. You can find the App ID in the Firebase console, on the General Settings page: ")
+          end
+        end
+
+        upload_token = get_upload_token(app_id, binary_path)
+        MAX_POLLING_RETRIES.times do
+          status = upload_status(upload_token, app_id)
+          if status == "SUCCESS"
+            UI.message("App Uploaded Successfully!")
```
tagboola, line 48: Change to "Uploaded APK successfully" to match Gradle plugin.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +223,46 @@ def self.is_firebasecmd_supported?(cmd)
...
           response = connection.get("#{PATH}#{app_id}") do |request|
-            request.headers["Authorization"] = "Bearer " + token
+            request.headers["Authorization"] = "Bearer " + auth_token
           end
-        rescue Faraday::ResourceNotFound => error
-          UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
-        rescue => error
-          UI.crash!("Failed to fetch app information: #{error.message}")
+        rescue Faraday::ResourceNotFound
+          UI.user_error!(FirebaseAppDistributionError::MISSING_APP_ID)
```
tagboola, line 233: I really like this! Looks much cleaner 🙂

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +223,46 @@ def self.is_firebasecmd_supported?(cmd)
...
+        UI.crash!(FirebaseAppDistributionError::GET_APP_ERROR)
+      end
+
+      def self.upload_status(app_token, app_id)
+        begin
+          response = connection.get("#{PATH}#{app_id}/upload_status/#{app_token}") do |request|
+            request.headers["Authorization"] = "Bearer " + auth_token
+          end
+        rescue Faraday::ResourceNotFound
+          UI.user_error!(FirebaseAppDistributionError::MISSING_APP_ID)
```
tagboola, line 261: I probably should update the docs to reflect this, but a 404 here really means there is no corresponding release, so I'm not sure you want to show the `MISSING_APP_ID` message here. I think you can remove the `Faraday::ResourceNotFound` rescue block and just rely on the rescue below.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +223,46 @@ def self.is_firebasecmd_supported?(cmd)
...
+        rescue
+          UI.crash!(FirebaseAppDistributionError::GET_APP_ERROR)
```
tagboola, line 263: This rescue block can be called for a number of reasons (permissions error, server error, throttling etc.). I think it would be better for debugging to pass the exception to the `UI.crash` method instead of passing a generic message i.e.

```
rescue => error
  puts UI.crash!(error)
end
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +223,46 @@ def self.is_firebasecmd_supported?(cmd)
...
+
+      def self.upload_binary(app_id, binary_path)
+        # check error on wrong binary path
+        connection.post("/app-binary-uploads?app_id=#{app_id}", File.open(binary_path).read) do |request|
+          request.headers["Authorization"] = "Bearer " + auth_token
+        end
+      rescue Faraday::ResourceNotFound
+        UI.user_error!(FirebaseAppDistributionError::MISSING_APP_ID)
+      rescue
+        UI.crash!(FirebaseAppDistributionError::GET_APP_ERROR)
```
tagboola, line 252: This rescue block can be called for a number of reasons (permissions error, server error, throttling etc.). I think it would be better for debugging to pass the exception to the `UI.crash` method instead of passing a generic message i.e.

```
rescue => error
  puts UI.crash!(error)
end
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +223,46 @@ def self.is_firebasecmd_supported?(cmd)
...
+            request.headers["Authorization"] = "Bearer " + auth_token
           end
-        rescue Faraday::ResourceNotFound => error
-          UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
-        rescue => error
-          UI.crash!("Failed to fetch app information: #{error.message}")
+        rescue Faraday::ResourceNotFound
+          UI.user_error!(FirebaseAppDistributionError::MISSING_APP_ID)
+        rescue
+          UI.crash!(FirebaseAppDistributionError::GET_APP_ERROR)
```
tagboola, line 235: This rescue block can be called for a number of reasons (permissions error, server error, throttling etc.). I think it would be better for debugging to pass the exception to the `UI.crash` method instead of passing a generic message i.e.

```
rescue => error
  puts UI.crash!(error)
end
```

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_exceptions.rb
```diff
@@ -0,0 +1,17 @@
+class FirebaseAppDistributionError < StandardError
```
tagboola, line 1: If all we will be doing is using the error messages in this class, it doesn't need to inherit from `StandardError`. Consider removing the inheritance and maying renaming  the class to `FirebaseAppDistributionErrorMessage` and the file to `firebase_app_distribution_error_message.rb`.

Or maybe we can just call it `ErrorMessage` so it's shorter 🤔 ? Up to you!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: I tried running the action without an app id and params.values does not ask me to input an app id. Instead we just get our error handling.

Screen Shot 2020-06-22 at 5 11 43 PM

alonsosalasinfante: https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/d0a48bf604920f07c622eb457ecd484cc7409100/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb#L129-L133 

Because the :app parameter is optional, the user won't be prompted to enter in the ID. If we set optional to False it will prompt the user if they don't specify it as a parameter but we decided to keep it optional because the App ID is optional if the platform is ios

tagboola: 👍

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +226,33 @@ def self.is_firebasecmd_supported?(cmd)
...
-          conn.adapter(Faraday.default_adapter)
-        end
+      def self.get_upload_token(app_id, binary_path)
+        binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
         begin
           response = connection.get("#{PATH}#{app_id}") do |request|
-            request.headers["Authorization"] = "Bearer " + token
+            request.headers["Authorization"] = "Bearer " + auth_token
           end
-        rescue Faraday::ResourceNotFound => error
```
tagboola, line 196: My mistake but we want to show the onboarding message if the `GET app` call returns a 404/ResourceNotFound exception, so you can add this rescue clause back.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,74 @@ class FirebaseAppDistributionAction < Action
...
+          app_id = params[:app]
+        else
+          if platform == :ios
+            archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
+            if archive_patharchive_path
+              app_id = findout_ios_app_id_from_archive(archive_path)
+            end
+          end
+          if app_id.nil?
+            app_id = UI.input("Your app's Firebase App ID. You can find the App ID in the Firebase console, on the General Settings page: ")
```
lfkellogg, line 39: Just to call it out: this actually doesn't match the previous behavior, which seems to be that it would call out to the Firebase CLI without setting `--app`. At that point the Firebase CLI would error out: https://github.com/firebase/firebase-tools/blob/master/src/commands/appdistribution-distribute.ts#L18. Am I right?

I'm OK with prompting the user for it here, I think we should just recognize that it's a change and we should probably run it by Alex.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,74 @@ class FirebaseAppDistributionAction < Action
...
+          end
+        end
+
+        upload_token = get_upload_token(app_id, binary_path)
+        if upload_status(upload_token, app_id) == "SUCCESS"
+          UI.message("This APK/IPA has been uploaded before. Skipping upload step.")
+        else
+          MAX_POLLING_RETRIES.times do
+            status = upload_status(upload_token, app_id)
```
lfkellogg, line 48: Getting the status again here is unnecessary the first time through. Can you avoid it?

One way would be to save off the `status` during the first call to `upload_status`, and then move this call to the end of the loop.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +226,35 @@ def self.is_firebasecmd_supported?(cmd)
...
+        binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
         begin
           response = connection.get("#{PATH}#{app_id}") do |request|
-            request.headers["Authorization"] = "Bearer " + token
+            request.headers["Authorization"] = "Bearer " + auth_token
           end
-        rescue Faraday::ResourceNotFound => error
-          UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
+        rescue Faraday::ResourceNotFound
+            UI.user_error!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
```
lfkellogg, line 236: Remove the extra 2-space indentation

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +226,35 @@ def self.is_firebasecmd_supported?(cmd)
...
         rescue => error
-          UI.crash!("Failed to fetch app information: #{error.message}")
+          UI.crash!(error)
```
lfkellogg, line 238: Like we talked about yesterday, instead of catching all unexpected errors and calling UI.crash!, you can just let the error go un-caught.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -167,43 +226,35 @@ def self.is_firebasecmd_supported?(cmd)
...
         end
-
         contact_email = response.body[:contactEmail]
-
         if contact_email.strip.empty?
-          UI.user_error!("We could not find a contact email for app #{app_id}. Please visit App Distribution within the Firebase Console to set one up.")
+          UI.user_error!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
```
lfkellogg, line 242: When there's no contact email, shouldn't the action stop executing? I think this just will just print an error to the screen and continue.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -0,0 +1,17 @@
+class ErrorMessage
```
lfkellogg, line 1: This can just be a `module` since it only contains constants.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
alonsosalasinfante: Sounds good, I reverted it to the previous behavior for now

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,70 @@ class FirebaseAppDistributionAction < Action
...
+        if params[:app] # Set app_id if it is specified as a parameter
+          app_id = params[:app]
+        elsif platform == :ios
+          archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
+          if archive_path
+            app_id = findout_ios_app_id_from_archive(archive_path)
+          end
+        end
+
+        upload_token = get_upload_token(app_id, binary_path)
```
lfkellogg, line 38: You now need to handle the case where `app_id` is null. Maybe print the UI.error! with the `MISSING_APP_ID` message and return?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
lfkellogg: The change didn't actually remove the unnecessary call to `upload_status`, but now you can just move this call to the end of the loop (line 54).

PR #60, opened 2020-06-26
Unit test

alonsosalasinfante:
Adds initial passing unit_tests to upload_binary, upload_status, and get_upload_token


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,56 @@ class FirebaseAppDistributionAction < Action
       FIREBASECMD_ACTION = "appdistribution:distribute".freeze
       BASE_URL = "https://firebaseappdistribution.googleapis.com"
       PATH = "/v1alpha/apps/"
+      MAX_POLLING_RETRIES = 60
```
pranavrajgopal, line 19: nit: I would rename this to be `POLLING_INTERVAL_SECONDS` (assuming that's the intention :)) just to be more explicit.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,56 @@ class FirebaseAppDistributionAction < Action
...
 
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
-        validate_app!(params[:app])
+        platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
+        binary_path = params[:ipa_path] || params[:apk_path]
+
+        if params[:app] # Set app_id if it is specified as a parameter
+          app_id = params[:app]
+        elsif platform == :ios
```
pranavrajgopal, line 31: Should there be an elseif for android? Or is that implicit with `params[:app]`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,56 @@ class FirebaseAppDistributionAction < Action
...
+          archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
+          if archive_path
```
pranavrajgopal, line 33: Could you rename `findout_ios_app_id_from_archive` to `get_ios_app_id_from_archive`? Feels more standard.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,16 +16,56 @@ class FirebaseAppDistributionAction < Action
...
+        @connection ||= Faraday.new(url: BASE_URL) do |conn|
+          conn.response(:json, parser_options: { symbolize_names: true })
+          conn.response(:raise_error) # raise_error middleware will run before the json middleware
+          conn.adapter(Faraday.default_adapter)
+        end
+      end
+
+      def self.auth_token
+        @auth_token ||= begin
+          client = Signet::OAuth2::Client.new(
```
pranavrajgopal, line 56: nit: create a constant for this url.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,214 @@
...
+    context 'when testing with invalid parameters' do
+      it 'should crash if given an invalid app_id' do
+        expected_path = "/v1alpha/apps/invalid_app_id/upload_status/app_token"
+        expect(temp_connection).to receive(:get).with(expected_path).and_raise(Faraday::ResourceNotFound.new("404"))
+        expect { Fastlane::Actions::FirebaseAppDistributionAction.upload_status("invalid_app_id", "app_token") }.to raise_error(ErrorMessage::INVALID_APP_ID)
+      end
     end
   end
 end
+
```
pranavrajgopal, line 96: nit: get rid of commented code

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,214 @@
 describe Fastlane::Actions::FirebaseAppDistributionAction do
   let(:fake_file) { StringIO.new }
-
-  before do
+  let(:temp_connection) { double("Connection") }
+  # These examples are copied directly from API responses. Maybe we should move them to another file?
+  let(:fake_get_app_response_valid) { double("Response", status: 200, body: { projectNumber: "1", appId: "1", platform: "android", bundleId: "1", contactEmail: "Hello@world.com" }) }
```
lfkellogg, line 5: I love all these test cases! Can you format them so they're a little easier to read?

Maybe something like:
```suggestion
  let(:fake_get_app_response_valid) do
    double("Response", status: 200, body: {
      projectNumber: "1",
      ...,
      ...,
    })
end
```

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,214 @@
...
+  let(:fake_upload_status_response_success) { double("Response", status: 200, body: { status: "SUCCESS", release: { distributedAt: "1970-01-01T00:00:00Z", id: "status_id", lastActivityAt: "1970-01-01T00:00:00Z", receivedAt: "1970-01-01T00:00:00Z", displayVersion: "1.0", buildVersion: "1" } }) }
+  let(:fake_upload_status_response_invalid_app_token) { double("Response", status: 200, body: { status: "ERROR", message: "Update your Gradle plugin to >=2.0.0 or your Firebase CLI to >=8.4.1 to use the faster, more reliable upload pipeline.", release: { distributedAt: "1970-01-01T00:00:00Z", lastActivityAt: "1970-01-01T00:00:00Z", receivedAt: "1970-01-01T00:00:00Z" } }) }
+  let(:fake_binary) { double("Binary") }
+
+  before(:each) do
+    allow(Fastlane::Actions::FirebaseAppDistributionAction).to receive(:connection).and_return(temp_connection)
+    allow(fake_binary).to receive(:read).and_return("Hello World")
+
+    allow(File).to receive(:open).and_return(fake_binary)
     allow(Tempfile).to receive(:new).and_return(fake_file)
```
lfkellogg, line 19: By the way I don't think you need Tempfile, do you? I'm guessing this was something for the old implementation/tests.

spec/firebase_app_distribution_action_spec.rb
lfkellogg: Also can you use different values for the app, bundle, and project IDs, so the person reading the test can see what's what? Feel free to use easy to read values like "test-project-id" or "fake-bundle-id"

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,214 @@
...
+      it 'should crash if the app has no contact email' do
+        expect(temp_connection).to receive(:get).with("/v1alpha/apps/app_id").and_return(fake_get_app_response_no_email)
+        expect { Fastlane::Actions::FirebaseAppDistributionAction.get_upload_token("app_id", "binary_path") }.to raise_error(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
       end
 
-      it "returns nil if param does not exist" do
-        params = {}
-        expect(action.flag_if_supplied('--debug', :debug, params)).to be_nil
+      it 'should crash if given an invalid app_id' do
+        expect(temp_connection).to receive(:get).with("/v1alpha/apps/invalid_app_id").and_raise(Faraday::ResourceNotFound.new("404"))
```
lfkellogg, line 41: Try to format these long lines so they are easier to read. For example `.with(...` and `.and_raise(...` can each drop to the next line.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,214 @@
...
       end
+    end
+  end
 
-      it "will use a path if supplied" do
-        params = {
-          testers_file: "/tmp/testers_file"
-        }
-        expect(action.testers_flag(params)).to eq("--testers-file /tmp/testers_file")
+  context '#upload_binary' do
```
lfkellogg, line 52: I think this is better as a `describe` because you're describing what this method does below.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
alonsosalasinfante: So this code is similar to the original 1.0 code:

https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/bb0dfea89fb4cb62ce8e63722ae98afc4234280c/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb#L21-L34

which also doesn't handle missing app_ids for android. The "if params[:app]" checks if the app id was passed in as a parameter, and we'll use it if so. Otherwise, we can try to find out the app id. I think in the future we'll have a get_app_id for android.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,84 +1,129 @@
...
-  let(:fake_upload_binary_response_valid) { double("Response", status: 202, body: { token: "projects/project_id/apps/app_id/releases/-/binaries/binary_hash" }) }
-  let(:fake_upload_status_response_success) { double("Response", status: 200, body: { status: "SUCCESS", release: { distributedAt: "1970-01-01T00:00:00Z", id: "status_id", lastActivityAt: "1970-01-01T00:00:00Z", receivedAt: "1970-01-01T00:00:00Z", displayVersion: "1.0", buildVersion: "1" } }) }
-  let(:fake_upload_status_response_invalid_app_token) { double("Response", status: 200, body: { status: "ERROR", message: "Update your Gradle plugin to >=2.0.0 or your Firebase CLI to >=8.4.1 to use the faster, more reliable upload pipeline.", release: { distributedAt: "1970-01-01T00:00:00Z", lastActivityAt: "1970-01-01T00:00:00Z", receivedAt: "1970-01-01T00:00:00Z" } }) }
+  let(:fake_connection) { double("Connection") }
   let(:fake_binary) { double("Binary") }
+  # These examples are copied directly from API responses. Maybe we should move them to another file?
```
lfkellogg, line 5: If you're concerned about these definitions taking up a large block at the beginning of the file, a couple options:
- Only define the fields needed for the test (for example, do you need the `distributedAt` values etc. in the upload response?)
- If the value is only needed for one test, move the `let` to inside that test's `define` or `context` block, below
- Similarly, if you only need it for one test, don't use a `let` at all, and just inline it in the test, e.g.:
```
....and_return({
  status: 200,
  body: {
    projectNumber: ...
})
```

After it's in a good spot, remove this comment.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,84 +1,129 @@
...
+    })
+  end
+  let(:fake_upload_status_response_success) do
+    double("Response", status: 200, body: {
+      status: "SUCCESS", release: { distributedAt: "1970-01-01T00:00:00Z", id: "status_id", lastActivityAt: "1970-01-01T00:00:00Z", receivedAt: "1970-01-01T00:00:00Z", displayVersion: "1.0", buildVersion: "1" }
+    })
+  end
+  let(:fake_upload_status_response_invalid_app_token) do
+    double("Response", status: 200, body: {
+      status: "ERROR", message: "Update your Gradle plugin to >=2.0.0 or your Firebase CLI to >=8.4.1 to use the faster, more reliable upload pipeline.", release: { distributedAt: "1970-01-01T00:00:00Z", lastActivityAt: "1970-01-01T00:00:00Z", receivedAt: "1970-01-01T00:00:00Z" }
```
lfkellogg, line 38: If you keep all these fields, please format the nested objects to make it more readable:
```
..., body: {
  status: ...,
  message: ...,
  release: {
    distributedAt: ...,
    ...
  },
}
```

You could also declare the `release` in its own variable and reference it here.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
pranavrajgopal: Sounds good!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -16,8 +16,9 @@ class FirebaseAppDistributionAction < Action
       FIREBASECMD_ACTION = "appdistribution:distribute".freeze
       BASE_URL = "https://firebaseappdistribution.googleapis.com"
       PATH = "/v1alpha/apps/"
+      TOKEN_PATH = "https://oauth2.googleapis.com/token"
```
jladieu, line 19: What kind of token?  `AUTH_TOKEN`?  Might consider giving a little more context in the name.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -16,8 +16,9 @@ class FirebaseAppDistributionAction < Action
...
       MAX_POLLING_RETRIES = 60
-      POLLING_INTERVAL_S = 2
+      POLLING_INTERVAL_SECONDS = 2
```
jladieu, line 21: Yes!  Great idea.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -31,7 +32,7 @@ def self.run(params)
         elsif platform == :ios
           archive_path = Actions.lane_context[SharedValues::XCODEBUILD_ARCHIVE]
           if archive_path
-            app_id = findout_ios_app_id_from_archive(archive_path)
+            app_id = get_ios_app_id_from_archive(archive_path)
```
jladieu, line 35: I like this rename :)

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
jladieu: Or maybe even `TOKEN_CREDENTIAL_URI` then it's easy to grep over to the way it gets used with the Oauth client

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,16 +210,22 @@ def self.is_firebasecmd_supported?(cmd)
       end
 
       def self.get_upload_token(app_id, binary_path)
-        binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
+        begin
+          binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
+        rescue Errno::ENOENT
+          UI.crash!(ErrorMessage::APK_NOT_FOUND)
```
jladieu, line 216: Might be worth including the `binary_path` in the error message here somehow, that way the customer can see what we tried.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,16 +210,22 @@ def self.is_firebasecmd_supported?(cmd)
...
+        end
+
         begin
           response = connection.get("#{PATH}#{app_id}") do |request|
             request.headers["Authorization"] = "Bearer " + auth_token
           end
         rescue Faraday::ResourceNotFound
-          UI.crash!("App Distribution could not find your app #{app_id}. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution")
+          UI.crash!(ErrorMessage::INVALID_APP_ID)
```
jladieu, line 224: Consider including the `app_id` you tried in the error message, or even the full path.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,16 +210,22 @@ def self.is_firebasecmd_supported?(cmd)
...
         end
+
         contact_email = response.body[:contactEmail]
-        if contact_email.strip.empty?
+        if contact_email.nil? || contact_email.strip.empty?
           UI.crash!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
```
jladieu, line 229: Consider including the original value of `contact_email` in a quoted string in the error message; this will help disambiguate whether it was just not provided, or a bunch of whitespace, etc.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -228,32 +235,44 @@ def self.upload_binary(app_id, binary_path)
         connection.post("/app-binary-uploads?app_id=#{app_id}", File.open(binary_path).read) do |request|
           request.headers["Authorization"] = "Bearer " + auth_token
         end
+      rescue Faraday::ResourceNotFound
+        UI.crash!(ErrorMessage::INVALID_APP_ID)
```
jladieu, line 239: Similar suggestions to before, if you can give the inputs that _caused_ the error in the error message, the customer might have an easier time self-resolving.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -228,32 +235,44 @@ def self.upload_binary(app_id, binary_path)
...
+      rescue Errno::ENOENT
+        UI.crash!(ErrorMessage::APK_NOT_FOUND)
       end
 
       def self.upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
-        status = upload_status(upload_token, app_id)
+        status = upload_status(app_id, upload_token)
```
jladieu, line 246: Howcome you decided to reorder the params?  No biggie just curious what your thoughts were.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -228,32 +235,44 @@ def self.upload_binary(app_id, binary_path)
...
             else
-              UI.message("Uploading the APK/IPA")
+              UI.message("Uploading the APK/IPA.")
               upload_binary(app_id, binary_path)
             end
-            status = upload_status(upload_token, app_id)
+            status = upload_status(app_id, upload_token)
+          end
+          if status != "SUCCESS"
+            UI.message("It took longer than expected to process your APK/IPA, please try again")
```
jladieu, line 264: Nit: `.` at the end

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -228,32 +235,44 @@ def self.upload_binary(app_id, binary_path)
...
-      def self.upload_status(app_token, app_id)
-        response = connection.get("#{PATH}#{app_id}/upload_status/#{app_token}") do |request|
-          request.headers["Authorization"] = "Bearer " + auth_token
+      def self.upload_status(app_id, app_token)
+        begin
+          response = connection.get("#{PATH}#{app_id}/upload_status/#{app_token}") do |request|
+            request.headers["Authorization"] = "Bearer " + auth_token
+          end
+        rescue Faraday::ResourceNotFound
+          UI.crash!(ErrorMessage::INVALID_APP_ID)
```
jladieu, line 275: Ditto to earlier suggestion about giving the failing value in the error message.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,131 @@
...
-      it "returns nil if param does not exist" do
-        params = {}
-        expect(action.flag_value_if_supplied('--token', :firebåse_cli_token, params)).to be_nil
+  describe '#get_upload_token' do
+    context 'when testing with valid parameters' do
+      it 'should make a GET call to the app endpoint and return the upload token' do
+        expect(fake_connection).to receive(:get)
+          .with("/v1alpha/apps/app_id")
+          .and_return(
+            double("Response", status: 200, body: {
```
jladieu, line 18: Consider instantiating and returning actual Response class of the type returned by the library you're using instead of returning a test double, which can't be guaranteed to behave like the code under test would.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,131 @@
 describe Fastlane::Actions::FirebaseAppDistributionAction do
   let(:fake_file) { StringIO.new }
+  let(:fake_connection) { double("Connection") }
+  let(:fake_binary) { double("Binary") }
 
-  before do
-    allow(Tempfile).to receive(:new).and_return(fake_file)
-    allow(fake_file).to receive(:unlink)
-    allow(fake_file).to receive(:path).and_return("/tmp/string")
+  before(:each) do
```
jladieu, line 6: I had some thoughts about mocking in tests in this PR https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/pull/61#pullrequestreview-440040774

I think the advice there dovetails nicely with what you've got here, but your specs all in all look very good and you guys can opt to completely disregard if you'd like.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,131 @@
...
-            firebase_cli_token: 'fake-token'
-        }
-        expect(action.flag_value_if_supplied('--token', :firebase_cli_token, params)).to eq("--token fake-token")
-      end
-
-      it "returns nil if param does not exist" do
-        params = {}
-        expect(action.flag_value_if_supplied('--token', :firebåse_cli_token, params)).to be_nil
+  describe '#get_upload_token' do
+    context 'when testing with valid parameters' do
```
jladieu, line 13: You can probably drop the `when testing` from the context message here.

But if possible I recommend avoiding using extra nesting with the `context` block because it can easily provide a way for you to accidentally introduce a [mystery guest](https://groups.google.com/a/google.com/forum/#!msg/testing-on-the-toilet/hci1Y-JM5g8/kheVY9WOBQAJ) by providing the perfect spot to start putting extra fixture code which becomes confusing.  If you can make an `it` clause that's terse enough, consider removing this extra `context` indentation entirely.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,131 @@
...
-        params = {
-            debug: true
-        }
-        expect(action.flag_if_supplied('--debug', :debug, params)).to eq("--debug")
-      end
-
-      it "returns nil if param does not exist" do
-        params = {}
-        expect(action.flag_if_supplied('--debug', :debug, params)).to be_nil
+    context 'when testing with invalid parameters' do
```
jladieu, line 32: Similar thought as earlier about potentially removing the extra `context` nesting here.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,131 @@
...
+      it 'should crash if the app has no contact email' do
+        expect(fake_connection).to receive(:get)
+          .with("/v1alpha/apps/app_id")
+          .and_return(
+            double("Response", status: 200, body: {
```
jladieu, line 37: Ditto on returning an actual `Response` instance here

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,131 @@
...
-        expect(action.testers_flag(params)).to eq("--testers-file /tmp/testers_file")
+      it 'should crash if given an invalid binary_path' do
+        expect(File).to receive(:open)
+          .with("invalid_binary_path")
+          .and_raise(Errno::ENOENT.new("file not found"))
+        expect { Fastlane::Actions::FirebaseAppDistributionAction.get_upload_token("app_id", "invalid_binary_path") }
+          .to raise_error(ErrorMessage::APK_NOT_FOUND)
       end
     end
+  end
```
jladieu, line 65: Great edge case testing here!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
alonsosalasinfante: Just to make it consistent with all the other method headers which have app_id as the first arg

spec/firebase_app_distribution_action_spec.rb
alonsosalasinfante: I'm a little confused by the mystery guest concept and unfortunately don't have access to the corp links. Are mystery guests when you have code say within the "when testing with valid parameters" portion that may affect the individual tests within that context?

jladieu: Updated the reference link to something publically shared.  Basically using context/let can provide an opportunity for 'surprises' to creep into your specs because the fixture setup happens far away from the actual test code.

PR #61, opened 2020-06-26
Post notes

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -207,6 +209,12 @@ def self.is_firebasecmd_supported?(cmd)
 
         true
       end
+      def self.post_notes(app_id, release_id, release_notes)
+       # payload = {releaseNotes: {releaseNotes: release_notes}}
+        connection.post("#{PATH}#{app_id}/releases/#{release_id}/notes", {releaseNotes: {releaseNotes: release_notes}}.to_json) do |request|
```
mannyjimenez0810, line 214: Was not sure if we should inline it or not so I commented out the initial approach.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -38,7 +38,9 @@ def self.run(params)
         if app_id.nil?
           UI.crash!(ErrorMessage::MISSING_APP_ID)
         end
-        upload(app_id, binary_path)
+        # Should be fine since we will have an error message if it goes over the limit of tries and every other time that it does not succed should end in some error so there shouldn't be any invalid input at this stage
+        # Release notes may need to check before passing it in that it isnt nil.
+        post_notes(app_id, upload(app_id, binary_path), params[:release_notes])
```
mannyjimenez0810, line 43: I put this as a comment, but upload will only return if it succeeds so we should be ok in terms of any invalid input here.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -246,16 +254,17 @@ def self.upload(app_id, binary_path)
               UI.message("Uploading the APK/IPA")
               upload_binary(app_id, binary_path)
             end
-            status = upload_status(upload_token, app_id)
+            status, release_id = upload_status(upload_token, app_id)
```
mannyjimenez0810, line 257: On cases of a failure the release_id is nil which is ok. In our other PR https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/pull/60 we have handled the case where the method would conclude not on a success if it reached the maximum tries.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -207,6 +209,12 @@ def self.is_firebasecmd_supported?(cmd)
 
         true
       end
+      def self.post_notes(app_id, release_id, release_notes)
+       # payload = {releaseNotes: {releaseNotes: release_notes}}
+        connection.post("#{PATH}#{app_id}/releases/#{release_id}/notes", {releaseNotes: {releaseNotes: release_notes}}.to_json) do |request|
+          request.headers["Authorization"] = "Bearer " + auth_token
+        end
```
mannyjimenez0810, line 216: Also I will be updating this with a message that the notes have been posted soon.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
lfkellogg: Would be good to reduce the length of this line by setting up the payload and/or the URL above

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -246,16 +259,17 @@ def self.upload(app_id, binary_path)
...
         end
+        release_id
       end
 
       def self.upload_status(app_token, app_id)
         response = connection.get("#{PATH}#{app_id}/upload_status/#{app_token}") do |request|
           request.headers["Authorization"] = "Bearer " + auth_token
         end
-        response.body[:status]
+        return response.body[:status], response.body[:release][:id]
```
lfkellogg, line 272: Clever! I'm not always a fan of returning tuples from methods, because it can make the interface kind of confusing. But in this case it's simple and cohesive enough that I like it.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
lfkellogg: Can you document this in a comment above the upload_status method? I think it would be good to document any method where the behavior, arguments, and return value aren't immediately clear from the signature.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,165 @@
...
-        testers_file: "/tmp/testers.txt",
-        release_notes_file: "/tmp/release_notes.txt",
-        firebase_cli_token: "fake-token",
-        debug: true
-      }
-      Fastlane::Actions::FirebaseAppDistributionAction.run(params)
+  context '#post_notes' do
+    context 'when testing with valid parameters' do
+      it 'should make a GET call to the app endpoint and return the upload token' do
+        expect(temp_connection).to receive(:post).with("/v1alpha/apps/app_id/releases/release_id/notes", "{\"releaseNotes\":{\"releaseNotes\":\"release_notes\"}}").and_return(fake_get_app_response_valid)
```
lfkellogg, line 23: This probably shouldn't return `fake_get_app_response_valid`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
lfkellogg: Pretty clever to inline the call to `upload`! In this case I would put it on its own line though, because the return value is unclear. I'd rather see: `release_id = upload(...` above so that I know we're passing a release ID to post_notes.

For example if the method was called `get_release_id` I think this would be fine. But that wouldn't make sense because `upload` is a much better description of what it does.

mannyjimenez0810: This works out since I will need the release ID later for enable access.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -243,6 +245,7 @@ def self.upload_binary(app_id, binary_path)
         end
       end
 
+      # When the binary is uploaded response body includes the release id and so that is stored for future upload methods.
```
lfkellogg, line 248: I think this would be better if it described the exact shape of the returned data. If you want to include the reason why, that's optional. But if you're going to include a doc string for the method, a quick summary of what it does is usually good to have first. So, for example:
```
# <what it does, e.g. "Uploads a binary...">
#
# <describe the return value, e.g. "Returns a tuple of..."><reason, e.g. "because we use the release ID to...">
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -245,7 +246,9 @@ def self.upload_binary(app_id, binary_path)
         end
       end
 
-      # When the binary is uploaded response body includes the release id and so that is stored for future upload methods.
+      # <"Uploads the binary">
+      #
+      # <"Returns the id of the release"> <explain "Only happens on a successful release">
```
lfkellogg, line 251: Oops! Sorry my comment was confusing but I didn't mean to actually include the `<>` and the quotation marks :)

Writing it in paragraph form would be great, for example:

```
# Uploads the binary
# 
# Returns the id of the release only on a successful release
```

mannyjimenez0810: I thought it was some ruby doc format, I will fix it now.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -38,7 +38,10 @@ def self.run(params)
         if app_id.nil?
           UI.crash!(ErrorMessage::MISSING_APP_ID)
         end
-        upload(app_id, binary_path)
+        # Should be fine since we will have an error message if it goes over the limit of tries and every other time that it does not succed should end in some error so there shouldn't be any invalid input at this stage
+        # Release notes may need to check before passing it in that it isnt nil.
```
lfkellogg, line 42: I think both of these comments can be removed, right?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -230,9 +245,12 @@ def self.upload_binary(app_id, binary_path)
         end
       end
 
+      # Uploads the binary
+      #
+      # Returns the id of the release"> <explain "Only happens on a successful release
```
lfkellogg, line 250: You've still got some extra `"`s and `<`s

Can you explain what happens if the release is not successful?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -246,16 +264,20 @@ def self.upload(app_id, binary_path)
...
-            status = upload_status(upload_token, app_id)
+            status, release_id = upload_status(upload_token, app_id)
           end
         end
+        release_id
       end
 
+      # Gets the upload status for the app release
+      #
+      # Returns the status and the id of the release of applicable
```
lfkellogg, line 275: typo: "of" -> "if"

Also, can you expand on what you mean by "if applicable"?

lfkellogg:
Just a couple minor things! Give this a rebase and I'll give it another look.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -208,6 +211,18 @@ def self.is_firebasecmd_supported?(cmd)
         true
       end
 
+      def self.post_notes(app_id, release_id, release_notes)
+        payload = { releaseNotes: { releaseNotes: release_notes } }
+        if release_notes.nil? || release_notes.empty?
+          UI.message("No Release notes passed in. Skipping this step")
```
jladieu, line 217: nit: maybe lowercase `Release` -> `release` and add a `.` after `Skipping this step`?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -208,6 +211,18 @@ def self.is_firebasecmd_supported?(cmd)
...
+          return
+        end
+        connection.post("#{PATH}#{app_id}/releases/#{release_id}/notes", payload.to_json) do |request|
```
jladieu, line 220: Two thoughts on this...
1) `PATH` is a pretty overloaded term.  I notice there's also a reference to an env variable $PATH.  I'm concerned of accidental overloading here.  Consider renaming our constant to `V1_ALPHA_APPS_PATH` to be more intention-revealing and also to disambiguate.
2) I noticed there are three (maybe more?) spots where we concatenate `#{PATH}{app_id}` -- maybe make this a helper method like
3) Consider dropping the trailing slash from the constant value and putting the slash right in the string concatenation, then it's clear where the path separator is between `PATH` and `app_id}
```
def self.v1_apps_path(app_id)
   "#{V1_ALPHA_APPS_PATH}/#{app_id}
end
```

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -246,16 +264,20 @@ def self.upload(app_id, binary_path)
...
+            status, release_id = upload_status(upload_token, app_id)
           end
         end
+        release_id
       end
 
+      # Gets the upload status for the app release
+      #
+      # Returns the status and the id of the release of applicable
       def self.upload_status(app_token, app_id)
```
jladieu, line 276: The name of this method looks outdated now...are you `uploading` a `status` or retrieving the `upload_status` of a thing? Also, now that it returns two variables, should it be updated to be called `get_upload_status_and_release_id`?

Last, I generally try to avoid returning more than one return value from a method as it can get unwieldly pretty quickly.   What combination of status & release_id can be returned?  Does a `release_id` always accompany a status of SUCCESS?  Is it always absent on anything else?  What other status values can be expect?  The definition of the expected results from this backend call might be worth fleshing out and documenting here by creating a more robust response object that encapsulates our expectations.

eg:

class ReleaseUploadStatusResponse
   attr_reader :status, :release_id
   def initialize(status, release_id)
      @status = stats
      @release_id = release_id
   end

   def successful?
      @status == 'SUCCESS'
   end

   # any other logic that might emerge from this refactoring
end
```

File: spec/firebase_app_distribution_action_spec.rb [hunk 15]
```diff
@@ -1,141 +1,147 @@
...
+  let(:fake_get_app_response_permission_denied) { double("Response", status: 403, body: { error: { code: 403, message: "The caller does not have permission", status: "PERMISSION_DENIED" } }) }
+  let(:fake_binary) { double("Binary") }
+
+  before(:each) do
+    allow(Fastlane::Actions::FirebaseAppDistributionAction).to receive(:connection).and_return(temp_connection)
+    allow(fake_binary).to receive(:read).and_return("Hello World")
+
+    allow(File).to receive(:open).and_return(fake_binary)
     allow(Tempfile).to receive(:new).and_return(fake_file)
     allow(fake_file).to receive(:unlink)
```
jladieu, line 16: I notice a lot of mocking in this spec setup which can be a sign of a [change detector test](https://sites.google.com/a/google.com/tater-tott/posts/2014/09/episode_350__change_detector_tests_considered_harmful)

I know it might be challenging to avoid mocks in this case because of all the filesystem interaction, but you might be able to remove one layer of mocking by creating your own version of a MockFile which will behave a little bit more like a real object than relying on you to aggressively mock every interaction out.  As long as the fake mostly follows the API you're pretending to be you might find it more reusable across specs and require fewer `allow(x).to receive` calls to get it ready for use.  You can flesh out your mock class as needed when you find more unmocked interactions, and you might find you can use it for various types of files in your spec.

EG:

```
class MockFile
   def initialize(path, contents)
     @path = path
     @contents = contents || ""
   end

   def read
      @contents 
   end

   def write(text)
     @contents << text # or overwrite, not sure what the behavior here should be
   end

   def path
     @path
   end
end
```

Also other projects commonly use something like [fakefs](https://github.com/fakefs/fakefs) to do this, but given this is an open source project the drawbacks of pulling in another dependency might outweigh the benefits of getting some of this support in your specs.  Probably better to remain lightweight.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,141 +1,147 @@
 describe Fastlane::Actions::FirebaseAppDistributionAction do
   let(:fake_file) { StringIO.new }
-
-  before do
+  let(:temp_connection) { double("Connection") }
+  let(:fake_get_app_response_valid) { double("Response", status: 200, body: { projectNumber: "1", appId: "1", platform: "android", bundleId: "1", contactEmail: "Hello@world.com" }) }
```
jladieu, line 4: I can't entirely tell from the diff where all these `let` variables get used, but consider inlining them in their corresponding test case if they are only used once.  It's easier to read specs when the setup is closer to the verification and can help you avoid some testing antipatterns like `The Mystery Guest` and `General Fixture` helpful ref: https://thoughtbot.com/blog/lets-not

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: The method is getting the status of the upload. On success release id has some value so it is not nil. The reason for returning two values is that to get the release id otherwise we would have to make another upload status call once we know it successfully uploaded. The response body from the backend can be seen in the documentation that Tunde made.

spec/firebase_app_distribution_action_spec.rb
mannyjimenez0810: A lot of the tests here are older tests that we will be deleting. I know you already looked at the unit test PR, but that has the format we will most likely use going forward.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: Thought three seems like a good idea. Will update that now to reflect this change.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -246,16 +264,20 @@ def self.upload(app_id, binary_path)
...
         response = connection.get("#{PATH}#{app_id}/upload_status/#{app_token}") do |request|
           request.headers["Authorization"] = "Bearer " + auth_token
         end
-        response.body[:status]
+        return response.body[:status], response.body[:release][:id]
```
jladieu, line 280: Piggybacking on an earlier comment about this, I had a similar initial thought that returning tuples isn't ideal...it might not be a far stretch to do something like this:

`return ReleaseUploadStatusResponse.new(response.body)` and then encapsulate examination of the JSON behind that class's interface (an example provided earlier).  That would allow this method not to think about fields on the JSON, and also help us avoid leaking the enums out to the calling code.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
jladieu: Based on the doc you shared, I think this API response format might lend itself well to being factored into a simple Ruby object created from the returned JSON (this is adjusted based on the doc you shared):

```

def ReleaseUploadStatusResponse
   def initialize(response_json_hash)
      @response_json_hash = response_json_hash
   end

   def status
      response_json_hash['status']
   end

   def success?
      status == 'SUCCESS' # || maybe other non-failure status count as success too?
   end

   def release_hash
      response_json_hash['release']
   end

   def release_id
      release_hash ? release_hash['id'] : nil
   end
end
 # etc...
```

Then you can use a nice higher level interface without having the client code have to be concerned with the format of the JSON response body.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -230,9 +245,12 @@ def self.upload_binary(app_id, binary_path)
         end
       end
 
+      # Uploads the binary
+      #
+      # Returns the id of the release"> <explain "Only happens on a successful release
       def self.upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
-        status = upload_status(upload_token, app_id)
+        status, release_id = upload_status(upload_token, app_id)
```
jladieu, line 253: If you encapsulate the response in a simple Ruby class as described in my other comment, this could become:

```
upload_status_response = upload_status(upload_token, app_id)

if upload_status_response.success?
   ... do stuff
elsif upload_status_response.in_progress?
   ... do other stuff
else
   ... 
end
```

No worries if you don't want to take this on at this time, but thought it might clean things up in this method a bit.

spec/firebase_app_distribution_action_spec.rb
jladieu: 👍

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: Thank you for the detailed comments and help. I have updated the code to reflect what you commented. Please let me know if I misunderstood anything.

mannyjimenez0810:
@googlebot I fixed it.

mannyjimenez0810:
@googlebot I fixed it.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,24 +214,36 @@ def self.is_firebasecmd_supported?(cmd)
...
+        end
+        UI.message("Release notes have been posted.")
+      end
+
       def self.get_upload_token(app_id, binary_path)
         begin
           binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
         rescue Errno::ENOENT
-          UI.crash!(ErrorMessage::APK_NOT_FOUND)
+          UI.crash!(ErrorMessage::APK_NOT_FOUND + binary_path)
```
lfkellogg, line 233: I don't think this will be formatted well. Do you want a `: ` in between the two strings? Maybe a format string would be good: `"#{ErrorMessage::APK_NOT_FOUND}: #{binary_path}"`

There are a couple of instances of this in this PR, so if you fix this one please fix them all.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,24 +214,36 @@ def self.is_firebasecmd_supported?(cmd)
...
         end
 
         begin
-          response = connection.get("#{PATH}#{app_id}") do |request|
+          response = connection.get(v1_apps_path(app_id).to_s) do |request|
```
lfkellogg, line 237: v1_apps_path returns a string so you shouldn't need the `to_s`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -236,45 +253,52 @@ def self.upload_binary(app_id, binary_path)
...
-        UI.crash!(ErrorMessage::INVALID_APP_ID)
+        UI.crash!(ErrorMessage::INVALID_APP_ID + app_id)
       rescue Errno::ENOENT
-        UI.crash!(ErrorMessage::APK_NOT_FOUND)
+        UI.crash!(ErrorMessage::APK_NOT_FOUND + binary_path)
       end
 
+      # Uploads the binary
+      #
+      # Returns the id of the release. Only happens on a successful release, on a fail release a messsage notifies the user.
```
lfkellogg, line 263: I think mentioning the return value in the failure case is more important than mentioning the message that's printed out. When we run out of retries, the return value should be nil, right?

It looks like we're also not handling this return value correctly when calling this method. Can you fix that too? I would also be fine with just crashing if that happens, in which case the return value does not matter or need to be handled.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -13,6 +13,6 @@ module ErrorMessage
   REFRESH_TOKEN_ERROR = "Could not generate credentials from the refresh token specified"
   GET_APP_ERROR = "App Distribution failed to fetch app information"
   APP_NOT_ONBOARDED_ERROR = "App Distribution not onboarded"
-  GET_APP_NO_CONTACT_EMAIL_ERROR = "App Distribution could not find a contact email associated with this app"
-  INVALID_APP_ID = "App Distribution could not find your app. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution"
+  GET_APP_NO_CONTACT_EMAIL_ERROR = "App Distribution could not find a contact email associated with this app. Contact Email: "
+  INVALID_APP_ID = "App Distribution could not find your app. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution. APP ID: "
```
lfkellogg, line 17: I think "App ID: " is better casing.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,24 +214,36 @@ def self.is_firebasecmd_supported?(cmd)
...
           end
         rescue Faraday::ResourceNotFound
-          UI.crash!(ErrorMessage::INVALID_APP_ID)
+          UI.crash!(ErrorMessage::INVALID_APP_ID + app_id)
         end
 
         contact_email = response.body[:contactEmail]
         if contact_email.nil? || contact_email.strip.empty?
-          UI.crash!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
+          UI.crash!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR + "\"#{contact_email}\"")
```
lfkellogg, line 246: I don't see any point in printing out the contact email if it's nil or empty.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -13,6 +13,6 @@ module ErrorMessage
   REFRESH_TOKEN_ERROR = "Could not generate credentials from the refresh token specified"
   GET_APP_ERROR = "App Distribution failed to fetch app information"
   APP_NOT_ONBOARDED_ERROR = "App Distribution not onboarded"
-  GET_APP_NO_CONTACT_EMAIL_ERROR = "App Distribution could not find a contact email associated with this app"
-  INVALID_APP_ID = "App Distribution could not find your app. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution"
+  GET_APP_NO_CONTACT_EMAIL_ERROR = "App Distribution could not find a contact email associated with this app. Contact Email: "
```
lfkellogg, line 16: I don't see any point in printing out the contact email if it's nil or empty.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -236,45 +253,52 @@ def self.upload_binary(app_id, binary_path)
...
+            UI.message("It took longer than expected to process your APK/IPA, please try again.")
           end
         end
+        upload_status_response.release_id
       end
 
-      def self.upload_status(app_id, app_token)
+      # Gets the upload status for the app release
+      #
+      # Returns the status of the release. On success the release id exists and is nil in all other cases.
```
lfkellogg, line 292: This is not true, it actually crashes if the status is not found, and never returns nil. Right?

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -10,90 +10,95 @@
...
 
-      it 'should crash if given an invalid binary_path' do
-        expect(File).to receive(:open)
-          .with("invalid_binary_path")
-          .and_raise(Errno::ENOENT.new("file not found"))
-        expect { Fastlane::Actions::FirebaseAppDistributionAction.upload_binary("app_id", "invalid_binary_path") }
-          .to raise_error(ErrorMessage::APK_NOT_FOUND)
-      end
+  describe '#post_notes' do
+    it 'should post the notes successfully' do
```
lfkellogg, line 97: Are there any other cases you should test here besides the success case?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: No if the status is not found it should not crash since this is just checking the status of the upload. So if the upload does not exist this will let us know if it is in progress, does not exist, or if it is successful. 

I have included a screenshot that prints out the body of the response in upload status for uploading a new apk. At first it is error since it hasn't been uploaded before, then inprogress, then succeeds at the third check. Only then is there a release id. 
Screen Shot 2020-07-01 at 10 45 12 AM

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: When it fails there should be an error message. I will update it so that it isn't a message but a user error that happens when the user runs out of tries.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/pull/60#discussion_r447722015  I believe we were addressing this comment by Josh. He states we should print the original contact email to disambiguate if it was an empty contact email or if it was actually not passed in.

jladieu: Yeah NBD on this -- Lee's probably right in that if it's an empty string (probably weird) it will be indistinguishable from nil in the output.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -236,45 +253,52 @@ def self.upload_binary(app_id, binary_path)
...
 
+      # Uploads the binary
+      #
+      # Returns the id of the release. Only happens on a successful release, on a fail it crashes.
       def self.upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
-        status = upload_status(app_id, upload_token)
-        if status == "SUCCESS"
+        upload_status_response = get_upload_status(app_id, upload_token)
+        if upload_status_response.success?
```
jladieu, line 267: Should this be `upload_status_response.already_uploaded?` (with a corresponding check for `ALREADY_UPLOADED` in the response class?)

jladieu: (I'm not sure I know the difference between SUCCESS vs. ALREADY_UPLOADED would occur)

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
```diff
@@ -0,0 +1,25 @@
+class UploadStatusResponse
```
jladieu, line 1: 👍

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
jladieu: Looks great :)  Thanks for taking this on.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: a success before we start the upload process means that the upload has already occurred and completed successfully. We could add "already_uploaded?", but it would be doing the same check as "success?"

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -236,45 +259,53 @@ def self.upload_binary(app_id, binary_path)
...
-        UI.crash!(ErrorMessage::INVALID_APP_ID)
+        UI.crash!("#{ErrorMessage::INVALID_APP_ID}: #{app_id}")
       rescue Errno::ENOENT
-        UI.crash!(ErrorMessage::APK_NOT_FOUND)
+        UI.crash!("#{ErrorMessage::APK_NOT_FOUND}: #{binary_path}")
       end
 
+      # Uploads the binary
+      #
+      # Returns the id of the release. Only happens on a successful release, on a fail it crashes.
```
lfkellogg, line 269: Please describe when `nil` can be returned.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -209,21 +217,36 @@ def self.is_firebasecmd_supported?(cmd)
         true
       end
 
+      def self.post_notes(app_id, release_id, release_notes)
+        payload = { releaseNotes: { releaseNotes: release_notes } }
+        if release_notes.nil? || release_notes.empty?
+          UI.success("No release notes passed in. Skipping this step.")
```
lfkellogg, line 223: This feels more neutral than a "success", maybe use UI.message?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -236,45 +259,53 @@ def self.upload_binary(app_id, binary_path)
...
+      #
+      # Returns the id of the release. Only happens on a successful release, on a fail it crashes.
       def self.upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
-        status = upload_status(app_id, upload_token)
-        if status == "SUCCESS"
-          UI.message("This APK/IPA has been uploaded before. Skipping upload step.")
+        upload_status_response = get_upload_status(app_id, upload_token)
+        if upload_status_response.success?
+          UI.success("This APK/IPA has been uploaded before. Skipping upload step.")
```
lfkellogg, line 274: This feels more neutral than a "success", maybe use UI.message?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -236,45 +259,53 @@ def self.upload_binary(app_id, binary_path)
...
+            return nil
           end
         end
+        upload_status_response.release_id
       end
 
-      def self.upload_status(app_id, app_token)
+      # Gets the upload status for the app release
+      #
+      # Returns the status of the release.
```
lfkellogg, line 299: "Returns the status of the release." is saying basically the same thing as "Gets the upload status for the app release" so you can remove it unless you want to elaborate.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,131 +1,193 @@
...
-              bundleId: "bundle_id",
-              contactEmail: "Hello@world.com"
-            })
-          )
-        upload_token = Fastlane::Actions::FirebaseAppDistributionAction.get_upload_token("app_id", "binary_path")
-        binary_hash = Digest::SHA256.hexdigest("Hello World")
-        expect(upload_token).to eq(CGI.escape("projects/project_number/apps/app_id/releases/-/binaries/#{binary_hash}"))
+    it 'should make a GET call to the app endpoint and return the upload token' do
+      stubs.get("/v1alpha/apps/app_id") do |env|
+        expect(env.url.path).to eq("/v1alpha/apps/app_id")
```
lfkellogg, line 33: This stubs thing is very cool. I'm curious: what value do you get from this expect? Won't this always be true, because if the url path was different, this block wouldn't get called?

Also, could we use `stubs.verify_stubbed_calls` to verify that the stub was called with the given url and body? See https://github.com/lostisland/faraday/blob/master/docs/adapters/testing.md

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,131 +1,193 @@
...
+          200,
+          {},
+          {}
+        ]
+      end
+      Fastlane::Actions::FirebaseAppDistributionAction.post_notes("app_id", "release_id", "release_notes")
+    end
+
+    it 'should not post if release notes are empty' do
+      expect(stubs).not_to(receive(:post))
```
lfkellogg, line 143: Should you be expecting `conn` to not receive post? My impression is that the stubs object is only used from the test, but I could be wrong.

spec/firebase_app_distribution_action_spec.rb
alonsosalasinfante: In this instance this will always be true. However, in cases like this: 

https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/945ec172480adb3b69f96018e5aa3fe3b2689ac2/spec/firebase_app_distribution_action_spec.rb#L90-L91

The env.url.path will be different than the stubs.post call because it does not include the additional args, only the path.

alonsosalasinfante: Also we do use the stubs.verify_stubbed_calls here: 

https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/945ec172480adb3b69f96018e5aa3fe3b2689ac2/spec/firebase_app_distribution_action_spec.rb#L25-L28

spec/firebase_app_distribution_action_spec.rb
alonsosalasinfante: Yes, fixed

spec/firebase_app_distribution_action_spec.rb
lfkellogg: Ah! I missed the after block. Thanks.

Even in the case like `/app-binary-uploads` I don't see the value in expecting the URL path. Even though it's different from the stubs.post call, it will still always be true because the block wouldn't execute otherwise.

PR #62, opened 2020-07-07
File support

mannyjimenez0810:
Adding in File support for the parameters. This allows users to pass in a file as a parameters for either for access or for release notes.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -215,6 +215,7 @@ module Fastlane
 
         true
       end
+
       def self.post_notes(app_id, release_id, release_notes)
         payload = { releaseNotes: { releaseNotes: release_notes } }
         if release_notes.nil? || release_notes.empty?
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -44,6 +44,7 @@ module Fastlane
           return
         end
         post_notes(app_id, release_id, params[:release_notes])
+        enable_access(app_id, release_id, params[:testers])
       ensure
         cleanup_tempfiles
       end
@@ -215,6 +216,16 @@ module Fastlane
         end
 
         true
+      end
+       def self.enable_access(app_id, release_id, emails)
+        #iterate through the string here and put it in the payload.
+        #payload = {emails: email}
+        #emails = email.gsub!(/\s+/, '').split(",")
+       payload = {emails: emails.gsub!(/\s+/, '').split(",")}
+        connection.post("#{PATH}#{app_id}/releases/#{release_id}/enable_access", payload.to_json) do |request|
+          request.headers["Authorization"] = "Bearer " + auth_token
+        end
+        UI.message("Successfully sent testers apk/ipa.")
       end
 
       def self.post_notes(app_id, release_id, release_notes)
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -43,10 +43,8 @@ module Fastlane
         if release_id.nil?
           return
         end
-        post_notes(app_id, release_id, params[:release_notes])
-        enable_access(app_id, release_id, params[:testers])
-      ensure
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,15 @@ module Fastlane
...
-        file_flag_if_supplied("--release-notes-file", "release_notes", params)
-      end
-
-      def file_flag_if_supplied(flag, param_name, params)
-        file = params["#{param_name}_file".to_sym]
-        file ||= file_for_contents(param_name.to_sym, params)
-
-        if file
-          return "#{flag} #{file}"
+      # Should we throw an exception if path is empty or not passed in? Check gradle plugin.
```
tagboola, line 9: We throw an exception in the gradle plugin if the file doesn't exist, so feel free to fail if that's the case.

https://github.com/FirebasePrivate/firebase-android-buildtools/blob/0d775e83e2686f87378e62f8716d1cff71c353d4/firebase-appdistribution/appdistribution-gradle/src/main/java/com/google/firebase/appdistribution/FirebaseAppDistributionOptions.java#L212

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -43,9 +43,8 @@ def self.run(params)
         if release_id.nil?
           return
         end
-        post_notes(app_id, release_id, params[:release_notes])
-      ensure
-        cleanup_tempfiles
+        release_notes = get_value_from_value_or_file(params[:release_notes], params[:release_notes_file])
```
tagboola, line 46: We should add a test for reading notes from a file.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
mannyjimenez0810: yea that is what we thought, but we wanted to be sure. I have updated it so that if there is a path that is not correct an error will be thrown to behave similar to the gradle plugin.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,16 @@ module Fastlane
...
+      def get_value_from_value_or_file(value, path)
+        if value.nil? || value.empty? && !path.nil? || !path.empty?
+          UI.message(!path.empty?)
+          if File.exist?(path)
+            return File.open(path).read
+          end
+          UI.crash!("#{ErrorMessage::APK_NOT_FOUND}: #{path}")
```
lfkellogg, line 14: For readability I think it's preferable to wrap a failure case like this in the actual failure condition. This has a few benefits: it makes the error self-documenting, it takes the failure case out of the main flow of the code, and in this case it also removes the need for the early `return`:
```suggestion
          if !File.exist?(path)
            UI.crash!("#{ErrorMessage::APK_NOT_FOUND}: #{path}")
          end
          File.open(path).read
```

Another option is the one-liner: `UI.crash!("#{ErrorMessage::APK_NOT_FOUND}: #{path}") unless File.exist?(path)` which can be nice but in this case might be a little too long to be readable.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -26,6 +26,23 @@
     stubs.verify_stubbed_calls
     Faraday.default_connection = nil
   end
+  describe '#get_value_from_value_or_file' do
```
lfkellogg, line 29: Should this be move to a different spec for just the helper module?

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -26,6 +26,23 @@
...
+    end
+
+    it 'should return the release notes from file' do
+      expect(File).to receive(:exist?).and_return(true)
+      expect(File).to receive(:open)
+        .with("release_notes_path")
+        .and_return(fake_binary)
+      expect(Fastlane::Actions::FirebaseAppDistributionAction.get_value_from_value_or_file("", "release_notes_path")).to eq("Hello World")
+    end
+    it 'should raise an error due to invalid release notes path ' do
```
lfkellogg, line 41: Put an empty newline above this line.

spec/firebase_app_distribution_action_spec.rb
mannyjimenez0810: We were thinking of moving it if the helper method gets bigger and requires more tests.

PR #63, opened 2020-07-08
Enable access

alonsosalasinfante:


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,23 @@ module Fastlane
...
-      end
-
-      ##
-      # always return a file for a given content
-      def file_for_contents(parameter_name, params)
-        if @tempfiles.nil?
-          @tempfiles = []
+      def string_to_array(string)
+        if string.nil?
+          return string
```
lfkellogg, line 22: Would an empty array be a more appropriate value to return?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,23 @@ module Fastlane
...
-        flag if params[param_name]
-      end
-
-      ##
-      # always return a file for a given content
-      def file_for_contents(parameter_name, params)
-        if @tempfiles.nil?
-          @tempfiles = []
+      def string_to_array(string)
+        if string.nil?
```
lfkellogg, line 21: Consider this Ruby-esque one-liner:
```suggestion
        return X if string.nil?
```

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -27,6 +27,87 @@
     Faraday.default_connection = nil
   end
 
+  describe '#get_value_from_value_or_file' do
+    it 'should return the release notes string' do
+      expect(Fastlane::Actions::FirebaseAppDistributionAction.get_value_from_value_or_file("Hello World", "")).to eq("Hello World")
```
lfkellogg, line 32: Try to make these lines shorter. For example, use an intermediate `result` variable, or maybe save off `Fastlane::Actions::FirebaseAppDistributionAction` into an `action` variable.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -27,6 +27,87 @@
     Faraday.default_connection = nil
   end
 
+  describe '#get_value_from_value_or_file' do
+    it 'should return the release notes string' do
```
lfkellogg, line 31: A few things:

- This method doesn't actually have to do with release notes, so the description shouldn't mention them. Instead it's just concerned with values and files in general.
- You can remove the "should" phrasing, i.e: "should return..." -> "returns..."
- The description should mention the specific case under test, in this case it's that the method returns the value *"when the value is defined"*

Please make these changes throughout.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -27,6 +27,87 @@
...
+      expect(Fastlane::Actions::FirebaseAppDistributionAction.get_value_from_value_or_file("Hello World", "")).to eq("Hello World")
+    end
+
+    it 'should return the release notes from file' do
+      expect(File).to receive(:open)
+        .with("release_notes_path")
+        .and_return(fake_binary)
+      expect(Fastlane::Actions::FirebaseAppDistributionAction.get_value_from_value_or_file("", "release_notes_path")).to eq("Hello World")
```
lfkellogg, line 39: This tests the case when the value is an empty string. Should you test the nil case as well?

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -27,6 +27,87 @@
...
+
+    it 'Should return an array with 3 testers' do
+      array = Fastlane::Actions::FirebaseAppDistributionAction.string_to_array("Tester1, Tester2, Tester3")
+      expect(array).to eq(["Tester1", "Tester2", "Tester3"])
+    end
+
+    it 'Should return nil if nil' do
+      array = Fastlane::Actions::FirebaseAppDistributionAction.string_to_array(nil)
+      expect(array).to eq(nil)
+    end
```
lfkellogg, line 65: Empty string would be another good test case

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -43,9 +43,15 @@ def self.run(params)
...
-        cleanup_tempfiles
+        release_notes = get_value_from_value_or_file(params[:release_notes], params[:release_notes_file])
+        testers = get_value_from_value_or_file(params[:testers], params[:testers_file])
+        groups = get_value_from_value_or_file(params[:groups], params[:groups_file])
+
+        post_notes(app_id, release_id, release_notes)
+        emails = string_to_array(testers)
+        group_ids = string_to_array(groups)
+
+        enable_access(app_id, release_id, emails, group_ids)
```
lfkellogg, line 54: I love these additions to the main run method. Super readable, nice work!!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -43,9 +43,15 @@ def self.run(params)
         if release_id.nil?
           return
```
tagboola, line 44: When is release_id ever `nil`? Should we be failing loudly if it is?

mannyjimenez0810: No this is when we try to many times and it doesn't upload successfully. This is for a graceful shutdown to give the user the fastlane summary at the end. See  https://google-fabric.slack.com/archives/C014P4EQMR7/p1593629261220000  for screenshots of the way it would shutdown with a return vs exit. A crash seemed like too much, but we can discuss if it is appropriate.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -217,6 +223,18 @@ def self.is_firebasecmd_supported?(cmd)
         true
       end
 
+      def self.enable_access(app_id, release_id, emails, group_ids)
+        if emails.nil? && group_ids.nil?
```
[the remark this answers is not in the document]
mannyjimenez0810: .present? Seems to give an error when I use it. I get an error message saying that the method was undefined. We will change the helper method to return nil if the string is empty to avoid any empty arrays.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,23 @@ module Fastlane
...
-      end
-
-      def file_flag_if_supplied(flag, param_name, params)
-        file = params["#{param_name}_file".to_sym]
-        file ||= file_for_contents(param_name.to_sym, params)
-
-        if file
-          return "#{flag} #{file}"
+      def get_value_from_value_or_file(value, path)
+        if (value.nil? || value.empty?) && !path.nil?
```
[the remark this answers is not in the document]
mannyjimenez0810: I mentioned this in a comment above, but present? seems to be giving me an error. This is the error: 
Screen Shot 2020-07-09 at 10 20 39 AM

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,23 @@ module Fastlane
...
-      def flag_if_supplied(flag, param_name, params)
-        flag if params[param_name]
-      end
-
-      ##
-      # always return a file for a given content
-      def file_for_contents(parameter_name, params)
-        if @tempfiles.nil?
-          @tempfiles = []
+      def string_to_array(string)
```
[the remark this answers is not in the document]
mannyjimenez0810: Makes sense. I have added in a comment for now to hopefully clarify that.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
mannyjimenez0810: I have decided to return nil in a response to another comment above this.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -6,60 +6,23 @@ module Fastlane
...
-      end
-
-      def cleanup_tempfiles
-        return if @tempfiles.nil?
-        @tempfiles.each(&:unlink)
-      end
-
-      def parse_plist(path)
-        CFPropertyList.native_types(CFPropertyList::List.new(file: path).value)
+        string_array = string.gsub(/\s+/, '').split(",")
```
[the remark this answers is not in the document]
mannyjimenez0810: We do not need to remove whitespace, but felt that it gave the customer more flexibility in terms of layout. Now they can have a string that is something like "Manny , Manny2 " rather than "Manny,Manny2" for readability. 

This method works with strings that have whitespace and those that don't. Also if we give the API a group or tester with a white space we will get a 400 status code.

spec/firebase_app_distribution_action_spec.rb
mannyjimenez0810: We have decided to go with the second option since it adds less lines of code and makes a lot of the lines shorter. We did not change all of the names to action since I believe we will have a PR for the refactoring and so these names may all change.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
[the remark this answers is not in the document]
lfkellogg: Ah, sorry turns out that a Rails only thing.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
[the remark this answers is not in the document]
lfkellogg: Yeah sorry this is a Rails-only thing. My mistake.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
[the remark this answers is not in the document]
lfkellogg: 👍

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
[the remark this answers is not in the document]
lfkellogg: > We do not need to remove whitespace, but felt that it gave the customer more flexibility in terms of layout. Now they can have a string that is something like "Manny , Manny2 " rather than "Manny,Manny2" for readability.

That makes sense 👍 

> This method works with strings that have whitespace and those that don't. Also if we give the API a group or tester with a white space we will get a 400 status code.

That's true, but as a reader if I look at the method name and the description I might try to use this in another spot to separate a comma-separated list of values where the values themselves might have spaces in them, like: "First Value, Second Value". And it wouldn't work for that. I think that's fine but you should make it clear to the reader, maybe in the method description.

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -3,6 +3,7 @@
   let(:fake_binary) { double("Binary") }
   let(:fake_auth_client) { double("auth_client") }
   let(:stubs) { Faraday::Adapter::Test::Stubs.new }
+  let(:action) { Fastlane::Actions::FirebaseAppDistributionAction }
```
lfkellogg, line 6: Nice 👍

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -28,28 +29,112 @@
...
+        .with("invalid_path")
+        .and_raise(Errno::ENOENT.new("file not found"))
+      expect { action.get_value_from_value_or_file(nil, "invalid_path") }
+        .to raise_error("#{ErrorMessage::INVALID_PATH}: invalid_path")
+    end
+  end
+
+  describe '#string_to_array' do
+    it 'returns an array when a string is passed in with no commas' do
+      array = Fastlane::Actions::FirebaseAppDistributionAction.string_to_array("string")
```
lfkellogg, line 73: Now that you have the `action` variable, please use it everywhere for consistency.

mannyjimenez0810: That was the plan, but since we were still refactoring I was unsure if it was worth making that change. Given the initial refactor PR then it makes sense to change it.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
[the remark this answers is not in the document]
mannyjimenez0810: That makes sense. I have added in an additional comment with an example hopefully to clarify that.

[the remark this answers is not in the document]
lfkellogg: Looks great 👍

PR #64, opened 2020-07-09
Refactor

alonsosalasinfante:
Moves api endpoint paths, and connection and auth_token objects to helper module

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -39,8 +39,15 @@ module Fastlane
         if release_id.nil?
           return
         end
+
         release_notes = get_value_from_value_or_file(params[:release_notes], params[:release_notes_file])
         fad_api_client.post_notes(app_id, release_id, release_notes)
+
+        testers = get_value_from_value_or_file(params[:testers], params[:testers_file])
+        groups = get_value_from_value_or_file(params[:groups], params[:groups_file])
+        emails = string_to_array(testers)
+        group_ids = string_to_array(groups)
+        fad_api_client.enable_access(app_id, release_id, emails, group_ids)
       end
 
       def self.description
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -5,6 +5,7 @@ require 'googleauth'
 require_relative '../helper/upload_status_response'
 require_relative '../helper/firebase_app_distribution_helper'
 require_relative '../helper/firebase_app_distribution_error_message'
+require_relative '../client/firebase_app_distribution_api_client'
 
 ## TODO: should always use a file underneath? I think so.
 ## How should we document the usage of release notes?
@@ -13,13 +14,13 @@ module Fastlane
     class FirebaseAppDistributionAction < Action
       DEFAULT_FIREBASE_CLI_PATH = `which firebase`
       FIREBASECMD_ACTION = "appdistribution:distribute".freeze
-      MAX_POLLING_RETRIES = 60
-      POLLING_INTERVAL_SECONDS = 2
 
       extend Helper::FirebaseAppDistributionHelper
+      # include Client::FirebaseAppDistributionApiClient # Is this needed? Should I use extend or include if so?
 
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
+        fad_api_client = Client::FirebaseAppDistributionApiClient
         platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
         binary_path = params[:ipa_path] || params[:apk_path]
 
@@ -35,12 +36,12 @@ module Fastlane
         if app_id.nil?
           UI.crash!(ErrorMessage::MISSING_APP_ID)
         end
-        release_id = upload(app_id, binary_path)
+        release_id = fad_api_client.upload(app_id, binary_path)
         if release_id.nil?
           return
         end
         release_notes = get_value_from_value_or_file(params[:release_notes], params[:release_notes_file])
-        post_notes(app_id, release_id, release_notes)
+        fad_api_client.post_notes(app_id, release_id, release_notes)
       end
 
       def self.description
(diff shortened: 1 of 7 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -15,8 +13,6 @@ module Actions
     class FirebaseAppDistributionAction < Action
       DEFAULT_FIREBASE_CLI_PATH = `which firebase`
```
tagboola, line 14: This isn't used anymore so feel free to remove it.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -1,11 +1,57 @@
 require 'fastlane_core/ui/ui'
-require 'cfpropertylist'
+require_relative '../actions/firebase_app_distribution_login'
 
 module Fastlane
   UI = FastlaneCore::UI unless Fastlane.const_defined?("UI")
 
   module Helper
     module FirebaseAppDistributionHelper
```
tagboola, line 8: I'm not sure you're gaining much by creating helpers to get the endpoint paths, connection and auth_token. The `FirebaseAppDistributionAction` still needs to how to use all those things e.g. call `connection.get(upload_status_path(app_id, app_token))` to get the upload status. I think it would be more useful to create a separate `FirebaseAppDistributionApiClient` class that will handle all the network calls. Then the the `run` method can just call `upload`, and the `FirebaseAppDistributionApiClient`, which contains all the necessary data, will make those calls.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -250,11 +219,11 @@ def self.get_upload_token(app_id, binary_path)
         if contact_email.nil? || contact_email.strip.empty?
           UI.crash!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
         end
-        return CGI.escape("projects/#{response.body[:projectNumber]}/apps/#{response.body[:appId]}/releases/-/binaries/#{binary_hash}")
+        return upload_token_format(response.body[:appId], response.body[:projectNumber], binary_hash)
       end
 
       def self.upload_binary(app_id, binary_path)
-        connection.post("/app-binary-uploads?app_id=#{app_id}", File.open(binary_path).read) do |request|
+        connection.post(binary_upload_path(app_id), File.open(binary_path).read) do |request|
```
lfkellogg, line 226: For the URL helper methods I think "*_url" would be a better naming scheme than "*_path" especially to distinguish from the file path.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
lfkellogg: Agree with Tunde's suggestion to create an API client class. So far the helper has been essentially a place to put some of the action code so that the action is fewer lines, but because there's no real difference between the two it's hard for the reader to understand what goes where, or why they are separate.

Instead, abstracting all of the network-related behavior into one class would be a clear separation of concerns: the client's job is to talk to the API, and that's what the action uses it for.

I do like the new URL helper methods though 👍 . I think it's nice to have them all next to each other.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
alonsosalasinfante: Can we remove the  configitem?

https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/bb0dfea89fb4cb62ce8e63722ae98afc4234280c/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb#L102-L118

tagboola: Ahh good point. I think we can remove this _if_ it doesn't break existing customers that have the `firebase_cli_path` set. We will have to test this scenario, and I'm not sure this is really in the scope of this PR so I'd so maybe hold off on this change 🙂

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -0,0 +1,146 @@
+require 'fastlane_core/ui/ui'
+require_relative '../actions/firebase_app_distribution_login'
+
+module Fastlane
+  module Client
+    module FirebaseAppDistributionApiClient # Class or module? I figured module sinze there is no state to track
```
tagboola, line 6: I think we should make this a class and make the methods below instance methods instead of class methods. At the moment there is technically not any configurable state, but I think it is likely that at some point it will be useful to have configurable state. For example, the caller could pass a flag to specify which login they would like to use as opposed to the client choosing the login method. Or the caller could pass a string specifying which environment to call i.e. prod, dev, test (this may be particularly useful for tests). Even though at the moment the `FirebaseAppDistributionAction` is the only caller, I think making this a class with instance methods (and maybe introducing configurable state to help with testing) is the better way to implement the api client.

alonsosalasinfante: That sounds good. Do you think I should keep the methods that we currently have as class method or change some/all of them to instance methods?

tagboola: Instance methods. Also any methods that you don't plan to expose to callers should be private or protected.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -0,0 +1,148 @@
...
+      def upload_status_url(app_id, app_token)
+        "#{v1_apps_url(app_id)}/upload_status/#{app_token}"
+      end
+
+      def upload_token_format(app_id, project_number, binary_hash)
+        CGI.escape("projects/#{project_number}/apps/#{app_id}/releases/-/binaries/#{binary_hash}")
+      end
+
+      def connection
+        @connection ||= Faraday.new(url: Helper::FirebaseAppDistributionHelper::BASE_URL) do |conn|
```
lfkellogg, line 125: BASE_URL has moved right? I'm guessing this would break if you tried to run it.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -0,0 +1,148 @@
...
+          conn.adapter(Faraday.default_adapter)
+        end
+      end
+
+      def auth_token
+        @auth_token ||= begin
+          client = Signet::OAuth2::Client.new(
+            token_credential_uri: TOKEN_CREDENTIAL_URI,
+            client_id: Fastlane::Actions::FirebaseAppDistributionLoginAction::CLIENT_ID,
+            client_secret: Fastlane::Actions::FirebaseAppDistributionLoginAction::CLIENT_SECRET,
```
lfkellogg, line 137: Do `CLIENT_SECRET` and `CLIENT_ID` still live in the Action? Should they be here in the client class?

alonsosalasinfante: They're in the LoginAction and also used there for authentication

https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/baeb25cd03c243a9a7446b5b804bf44ca8e7ffd8/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb#L11-L17

Since our auth/login is going to change, where do you think should be the best place for this?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
alonsosalasinfante: nice catch

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -61,7 +40,7 @@
           }
         ]
       end
-      upload_token = Fastlane::Actions::FirebaseAppDistributionAction.get_upload_token("app_id", "binary_path")
+      upload_token = api_client.get_upload_token("app_id", "binary_path")
```
tagboola, line 43: It's not immediately clear to me when reading this test that the path `"binary_path"` refers to the fake file that you stubbed out above. It also isn't immediately clear that the content of the fake file is "Hello World". I think an easy way to make this more readable is to use descriptive variable names. Maybe a `let` helper for both i.e. 

```
let(:fake_binary_path) { "binary_path" }
let(:fake_binary_contents) { "Hello World" } 
```

Another thing you can do is make it more visible by placing it closer to the test code inside the `it` block (like what you do in the "should crash if given an invalid binary_path" test)  i.e.

```
allow(fake_binary).to receive(:read).and_return("Hello World")
allow(File).to receive(:open)
        .with("binary_path")
        .and_return(fake_binary)
```

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: Ahh that makes sense. I wonder if logging in should be handled by the API? I don't think it's necessary to make that decision right now but definitely something to think about as we add more ways to authenticate.

lfkellogg: Ah, I see. Agreed, feel free to leave for now if you want.

PR #81, opened 2020-07-20
Updates upload method description and adds specs

alonsosalasinfante:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -72,10 +72,16 @@ def upload_binary(app_id, binary_path)
...
 
-      # Uploads the binary
+      # Uploads the binary file if it has not already been uploaded
+      # Takes at least POLLING_INTERVAL_SECONDS between retrying uploads
       #
-      # Returns the release_id on a successful release.
-      # Returns nil if unable to upload.
+      # args
+      #   app_id - Firebase App ID
+      #   binary_path - Absolute path to you app's apk/ipa file
```
lfkellogg, line 80: `you` -> `your` :)

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -72,10 +72,16 @@ def upload_binary(app_id, binary_path)
...
+      #
+      # Returns the release_id on a successful release, otherwise returns nil.
+      #
+      # Throws an error if the number of polling attempts exceeds MAX_POLLING_RETRIES
```
lfkellogg, line 84: I love this method description format you used: it really works for a method like this that could use some explanation. Did you come up with this format yourself or adapt it from somewhere?

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+    end
+
+    it 'skips the upload step if the binary has already been uploaded' do
+      expect(api_client).to receive(:get_upload_token)
+        .with("app_id", fake_binary_path)
+        .and_return("upload_token")
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_success)
+      expect(api_client).not_to(receive(:get_upload_status))
```
lfkellogg, line 158: Should this be expecting `:upload_binary` to not be called?

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+        .with("app_id", fake_binary_path)
+        .and_return("upload_token")
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error)
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path)
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_success)
```
lfkellogg, line 175: Since you're expecting :get_upload_status to get called twice, can you expect the order of the calls as well?

Also, I think this might be a thing: `expect(thing).to receive(:method).and_return(first_call_result, second_call_result, third_call_result)`

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+
+    it 'uploads the app binary then returns the release_id' do
+      expect(api_client).to receive(:get_upload_token)
+        .with("app_id", fake_binary_path)
+        .and_return("upload_token")
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error)
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path)
```
lfkellogg, line 172: Can you expect that this is only called once? In general consider if there are cases where it makes sense to expect the number of times things are called.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
   end
 
   describe '#upload' do
-    # Empty for now
+    let(:max_polling_retries) { Fastlane::Client::FirebaseAppDistributionApiClient::MAX_POLLING_RETRIES }
+    let(:polling_interval_seconds) { Fastlane::Client::FirebaseAppDistributionApiClient::POLLING_INTERVAL_SECONDS }
```
lfkellogg, line 131: Is this used?

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_success)
+
+      release_id = api_client.upload("app_id", fake_binary_path)
+      expect(release_id).to eq("release_id")
+    end
+
+    it 'attempts to upload MAX_POLLING_RETRIES times' do
+      stub_const("Fastlane::Client::FirebaseAppDistributionApiClient::POLLING_INTERVAL_SECONDS", 0)
```
lfkellogg, line 182: Instead of setting this to zero locally to each test, can you do it globally for this test class? What I'm afraid of is someone adding a test, forgetting to set this, and accidentally adding a test that takes several seconds or even minutes to run.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+      expect(api_client).to receive(:get_upload_token)
+        .with("app_id", fake_binary_path)
+        .and_return("upload_token")
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error)
+      max_polling_retries.times do
```
lfkellogg, line 189: I think this might be built into rspec: `expect(thing).to receive(:method).with(args).exactly(3).times`

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+      stub_const("Fastlane::Client::FirebaseAppDistributionApiClient::POLLING_INTERVAL_SECONDS", 0)
+      expect(api_client).to receive(:get_upload_token)
+        .with("app_id", fake_binary_path)
+        .and_return("upload_token")
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error)
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path)
+      (max_polling_retries / 2).times do
```
lfkellogg, line 211: I like the idea of making the test completely independent of the actual MAX_POLLING_RETRIES value. Consider stubbing it like you did with POLLING_INTERVAL_SECONDS, and then you wouldn't have to do this type of thing, you could just choose a number of times to poll.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
+    end
+    let(:upload_status_response_error) do
+      UploadStatusResponse.new(
+        { status: "ERROR",
+          release: {} }
+      )
+    end
+
+    it 'skips the upload step if the binary has already been uploaded' do
+      expect(api_client).to receive(:get_upload_token)
```
tagboola, line 152: It looks like every spec is setting this expectation. Consider placing it in a before filter.

spec/firebase_app_distribution_api_client_spec.rb
tagboola: +1 Nice find Lee!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
alonsosalasinfante: Not really, I couldn't find a general format that most people use for ruby unless they're using a docwriter so I just went with this. I'm planning on writing all the other method descriptions like this

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: Nice! I was wondering if there was a way of writing them like that. I believe they are expectations are ordered by how they're written

alonsosalasinfante: Nevermind, they're not but you can make them: https://til.hashrocket.com/posts/ex4n9iqwnw-message-order-constraint-in-rspec

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: Nope, forgot to take them out

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: That makes sense, I'll stub it for just the #upload tests

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: Good catch!

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: in this test I'm testing that the upload method does not try to upload again after seeing that the upload status is a success. I'll add comments to make it clearer

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: nice!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
lfkellogg: Sounds good - and feel free to pick and choose which pieces you include, for example leaving out the args part if it's self explanatory

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,92 @@
...
+
+    it 'uploads the app binary then returns the release_id' do
+      # return an error then a success after being uploaded
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error, upload_status_response_success).ordered
+
+      # upload_binary should only be called once
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path).ordered
```
lfkellogg, line 172: The comment says "should only be called once" but this expectation doesn't assert that, right?

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,92 @@
...
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error).ordered
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path).ordered
+      # return in_progress for a couple polls
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(UploadStatusResponse.new({ status: "IN_PROGRESS", release: {} }))
+        .exactly(max_polling_retries / 2).ordered
```
lfkellogg, line 208: Rather than using 60 and n/2 for these counts, since you control them now consider using something smaller and easier to reason about, like a hard-coded 3 and 2.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,92 @@
...
+      max_polling_retries = 60
+      stub_const("Fastlane::Client::FirebaseAppDistributionApiClient::MAX_POLLING_RETRIES", max_polling_retries)
+
+      expect(api_client).to receive(:get_upload_status)
+        .with("app_id", "upload_token")
+        .and_return(upload_status_response_error)
+        .at_least(:once).ordered
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path)
+        .exactly(max_polling_retries).times.ordered
```
lfkellogg, line 188: In general I think the use of `.ordered` is kind of hard to reason about. If you don't specifically need to assert that the methods were called in a particular order, I would just leave it out. I think expecting the number of calls should cover your test cases.

And in the case of you wanting a mock to return different values in a particular order, `.and_return(result_1, result_2)` gets you there and you don't need `.ordered`.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,92 @@
...
+
+      # Expect a call to get_upload_token every time
+      expect(api_client).to receive(:get_upload_token)
+        .with("app_id", fake_binary_path)
+        .and_return("upload_token").ordered
+    end
+
+    it 'skips the upload step if the binary has already been uploaded' do
+      # upload should not attempt to upload the binary at all
+      expect(api_client).not_to(receive(:upload_binary))
```
tagboola, line 155: nit: Prefer `expect(api_client).to_not receive(:upload_binary)`

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,92 @@
...
+
+      # upload_binary should only be called once
+      expect(api_client).to receive(:upload_binary)
+        .with("app_id", fake_binary_path).ordered
+
+      release_id = api_client.upload("app_id", fake_binary_path)
+      expect(release_id).to eq("release_id")
+    end
+
+    it 'attempts to upload MAX_POLLING_RETRIES times' do
```
tagboola, line 178: I don't think we should be attempting to upload the binary `MAX_POLLING_RETRIES` times. After uploading the binary, we should be polling for the upload status `MAX_POLLING_RETRIES` times.

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: Yup, I forgot to append .at_most(:once)

spec/firebase_app_distribution_api_client_spec.rb
alonsosalasinfante: That's right, I'll change the current test and method description

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -128,7 +127,99 @@
...
-    # Empty for now
+    let(:max_polling_retries) { Fastlane::Client::FirebaseAppDistributionApiClient::MAX_POLLING_RETRIES }
+    let(:polling_interval_seconds) { Fastlane::Client::FirebaseAppDistributionApiClient::POLLING_INTERVAL_SECONDS }
+    let(:upload_status_response_success) do
+      UploadStatusResponse.new(
+        { status: "SUCCESS",
+          release: { id: "release_id" } }
+      )
+    end
+    let(:upload_status_response_in_progress) do
```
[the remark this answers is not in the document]
alonsosalasinfante: Left it in since it's used in two specs

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -144,12 +238,12 @@
     end
 
     it 'does not post when the release notes are empty' do
-      expect(conn).not_to(receive(:post))
+      expect(conn).to_not(receive(:post))
```
tagboola, line 241: nit: Remove the parentheses for the `no_not` method so it looks like  `expect(conn).to_not receive(:post)`

alonsosalasinfante: It looks like rubocop prefers parentheses:

```
spec/firebase_app_distribution_api_client_spec.rb:241:20: C: [Corrected] Use parentheses for method calls with arguments.
      expect(conn).to_not receive(:post)
```

PR #83, opened 2020-07-22
Adds method decription, refactors error handling, and adds specs

alonsosalasinfante:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -9,19 +9,31 @@ class FirebaseAppDistributionApiClient
...
-        begin
         payload = { emails: emails, groupIds: group_ids }
-        connection.post(enable_access_url(app_id, release_id), payload.to_json) do |request|
-          request.headers["Authorization"] = "Bearer " + auth_token
+        begin
+          connection.post(enable_access_url(app_id, release_id), payload.to_json) do |request|
+            request.headers["Authorization"] = "Bearer " + auth_token
+          end
+        rescue Faraday::ResourceNotFound
+          UI.user_error!("#{ErrorMessage::INVALID_APP_ID}: #{app_id}")
```
lfkellogg, line 33: This change makes me realize that these should probably `.crash!` instead of user_error. It's weird to have it print an error and then immediately afterwards print the success message on line 37.

If you make that change please update the doc comment as well.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -9,19 +9,31 @@ class FirebaseAppDistributionApiClient
...
+      #   app_id - Firebase App ID
+      #   release_id - App release ID, returned by upload_status endpoint
+      #   emails - String array of app testers' email addresses
+      #   group_ids - String array of Firebase tester group IDs
+      #
+      # Throws a user_error if app_id, emails, or group_ids are invalid
       def enable_access(app_id, release_id, emails, group_ids)
         if emails.nil? && group_ids.nil?
           UI.message("No testers passed in. Skipping this step")
           return
```
lfkellogg, line 25: Your comment has made me realize that in this case, the user will not see the success message ("App Distribution upload finished successfully"), since it is printed at the bottom of this method. I think that really belongs in the run method, could you move it there?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
lfkellogg: Oh wait, do these actually cause a crash? Your tests make it seem like they do. If so please disregard my comment.

tagboola: Technically this can fail if the app or release isn't found. I think the error message should be something like:
```
App distribution failed to fetch release with id <release_id> 
```

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -205,7 +205,7 @@
       api_client.enable_access("app_id", "release_id", ["testers"], ["groups"])
     end
 
-    it 'posts when groupIds are defined and tester emails is nil' do
+    it 'posts when group_ids are defined and tester emails is nil' do
```
tagboola, line 208: Is rubocop making you change this? If not, you can leave it as `groupIds` since that's how it's passed in the payload.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -9,19 +9,31 @@ class FirebaseAppDistributionApiClient
...
+      #
+      # args
+      #   app_id - Firebase App ID
+      #   release_id - App release ID, returned by upload_status endpoint
+      #   emails - String array of app testers' email addresses
+      #   group_ids - String array of Firebase tester group IDs
+      #
+      # Throws a user_error if app_id, emails, or group_ids are invalid
       def enable_access(app_id, release_id, emails, group_ids)
         if emails.nil? && group_ids.nil?
```
tagboola, line 23: Consider checking if `emails` and `groups` are empty as well (and adding a test 🙂)

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
alonsosalasinfante: @lfkellogg user_error does cause a crash. @tagboola I added this functionality to the post_notes pr but I can't to this method because an invalid release_id returns a client error just like invalid emails/goup_ids. How should we handle it in this case? I think that since the user should not be specifying the release_id (since the plugin just gets that from upload_status), it's not as helpful to catch this and throw it as a user_error. 
However, I do think that an invalid release_id would be a reasonable exception to catch -  will the backend eventually have functionality to differentiate between invalid release_ids vs emails/groupids?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
alonsosalasinfante: Done

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: Ahh okay. So an invalid release_id doesn't return a 404? If that's the case I'm fine with leaving it the way it is 🙂

PR #86, opened 2020-07-23
Adds post_notes description, changes error handling

alonsosalasinfante:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -40,7 +49,7 @@ def post_notes(app_id, release_id, release_notes)
             request.headers["Authorization"] = "Bearer " + @auth_token
           end
         rescue Faraday::ResourceNotFound
-          UI.crash!("#{ErrorMessage::INVALID_APP_ID}: #{app_id}")
+          UI.user_error!("#{ErrorMessage::INVALID_APP_ID}: #{app_id}")
```
tagboola, line 52: Technically this can fail if the app or release isn't found. I think the error message should be something like:
```
App distribution failed to fetch release with id <release_id> 
``

PR #94, opened 2020-07-29
Adds error handling for upload errors, adds a new spec to for upload_…

alonsosalasinfante:
…binary

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
@@ -107,25 +107,27 @@ module Fastlane
       #
       # Returns the release_id on a successful release, otherwise returns nil.
       #
-      # Throws an error if the number of polling retries exceeds MAX_POLLING_RETRIES
+      # Throws a UI error if the number of polling retries exceeds MAX_POLLING_RETRIES
+      # or if it was not able to upload the binary
       def upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
         upload_status_response = get_upload_status(app_id, upload_token)
         if upload_status_response.success?
           UI.success("This APK/IPA has been uploaded before. Skipping upload step.")
         else
-          UI.message("This APK has not been uploaded before.")
+          UI.message("Uploading the APK/IPA.")
+          upload_binary(app_id, binary_path)
           MAX_POLLING_RETRIES.times do
+            upload_status_response = get_upload_status(app_id, upload_token)
             if upload_status_response.success?
               UI.success("Uploaded APK/IPA Successfully!")
               break
             elsif upload_status_response.in_progress?
               sleep(POLLING_INTERVAL_SECONDS)
             else
-              UI.message("Uploading the APK/IPA.")
-              upload_binary(app_id, binary_path)
+              UI.error("#{ErrorMessage::UPLOAD_APK_ERROR}: #{upload_status_response.message}")
+              return nil
             end
-            upload_status_response = get_upload_status(app_id, upload_token)
           end
           unless upload_status_response.success?
             UI.error("It took longer than expected to process your APK/IPA, please try again.")
--- a/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
@@ -108,7 +108,7 @@ module Fastlane
       # Returns the release_id on a successful release, otherwise returns nil.
       #
       # Throws a UI error if the number of polling retries exceeds MAX_POLLING_RETRIES
-      # or if it was not able to upload the binary
+      # Crashes if not able to upload the binary
       def upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
         upload_status_response = get_upload_status(app_id, upload_token)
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -107,25 +107,27 @@ def upload_binary(app_id, binary_path)
...
-      # Throws an error if the number of polling retries exceeds MAX_POLLING_RETRIES
+      # Throws a UI error if the number of polling retries exceeds MAX_POLLING_RETRIES
+      # or if it was not able to upload the binary
       def upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
         upload_status_response = get_upload_status(app_id, upload_token)
         if upload_status_response.success?
           UI.success("This APK/IPA has been uploaded before. Skipping upload step.")
         else
-          UI.message("This APK has not been uploaded before.")
```
tagboola, line 117: We should leave this log message and change it to:

‘This APK/IPA has not been uploaded before’

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -107,25 +107,27 @@ def upload_binary(app_id, binary_path)
...
+            upload_status_response = get_upload_status(app_id, upload_token)
             if upload_status_response.success?
               UI.success("Uploaded APK/IPA Successfully!")
               break
             elsif upload_status_response.in_progress?
               sleep(POLLING_INTERVAL_SECONDS)
             else
-              UI.message("Uploading the APK/IPA.")
-              upload_binary(app_id, binary_path)
+              UI.error("#{ErrorMessage::UPLOAD_APK_ERROR}: #{upload_status_response.message}")
```
tagboola, line 128: Do we need to return nil here? Should we just fail loudly with a `UI.error_message`?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -107,25 +107,27 @@ def upload_binary(app_id, binary_path)
...
+      # or if it was not able to upload the binary
       def upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
         upload_status_response = get_upload_status(app_id, upload_token)
         if upload_status_response.success?
           UI.success("This APK/IPA has been uploaded before. Skipping upload step.")
         else
-          UI.message("This APK has not been uploaded before.")
+          UI.message("Uploading the APK/IPA.")
+          upload_binary(app_id, binary_path)
```
lfkellogg, line 119: The one case I see this logic having issues is if the upload times out and they try again, and the previous upload is still in progress. In that case, we would try to upload again here, and we would get an error from the server because the binary still exists (I assume) and crash.

I think ideally in that scenario we would skip the second upload and enter the loop to keep checking the status.

tagboola: +1 good catch Lee!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: Consider handling the `status=ALREADY_UPLOADED` case. Even though we currently aren't returning `status=ALREADY_UPLOADED` on the backend, in the future we could and if we did this would incorrectly show an error message.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -107,25 +107,33 @@ def upload_binary(app_id, binary_path)
...
       def upload(app_id, binary_path)
         upload_token = get_upload_token(app_id, binary_path)
         upload_status_response = get_upload_status(app_id, upload_token)
-        if upload_status_response.success?
+        if upload_status_response.success? || upload_status_response.already_uploaded?
           UI.success("This APK/IPA has been uploaded before. Skipping upload step.")
         else
-          UI.message("This APK has not been uploaded before.")
+          UI.message("This APK/IPA has not been uploaded before")
+          UI.message("Uploading the APK/IPA.")
```
lfkellogg, line 119: Seems like these messages should be inside the `unless`

tagboola: I think it's _okay_ if these are outside of the `unless` statement. For example: A developer uploads their binary, then for some reason they kill the current process. If they rerun it and the binary is still being processed in which case `upload_status_response.in_progress?` is `true`, if the message statements were inside of the `unless` statement it wouldn't display any messages.

Even though we _technically_ aren't uploading the binary again, I think it's better to show the messages tor provide information the developer. @lfkellogg thoughts?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -107,25 +107,33 @@ def upload_binary(app_id, binary_path)
...
+            if upload_status_response.success? || upload_status_response.already_uploaded?
               UI.success("Uploaded APK/IPA Successfully!")
               break
             elsif upload_status_response.in_progress?
               sleep(POLLING_INTERVAL_SECONDS)
             else
-              UI.message("Uploading the APK/IPA.")
-              upload_binary(app_id, binary_path)
+              if !upload_status_response.message.nil?
+                UI.crash!("#{ErrorMessage::UPLOAD_APK_ERROR}: #{upload_status_response.message}")
```
tagboola, line 132: Consider using `user_error!` since this an expected user error e.g. an incorrectly formatted file.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -218,6 +225,28 @@
       release_id = api_client.upload("app_id", fake_binary_path)
       expect(release_id).to eq("release_id")
     end
+
```
tagboola, line 228: Consider adding a test for the scenario where the initial call to get the upload_status returns a response with status equal to `IN_PROGRESS`. For this, the test would make sure that upload is **not** called.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
lfkellogg: Yeah that makes sense, thanks. I think it's fine!

PR #107, opened 2020-07-30
Adding in ios app id support

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -23,6 +24,10 @@ def string_to_array(string)
         return string_array
       end
 
+      def parse_plist(path)
+        CFPropertyList.native_types(CFPropertyList::List.new(file: path).value)
+      end
+
       def get_ios_app_id_from_archive(path)
```
tagboola, line 31: Do we have a test for this? It would be great if this bug was caught by a failing test 🙂

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -23,8 +24,13 @@ def string_to_array(string)
...
         app_path = parse_plist("#{path}/Info.plist")["ApplicationProperties"]["ApplicationPath"]
+        puts(app_path)
```
lfkellogg, line 33: Remove this

spec/firebase_app_distribution_helper_spec.rb
```diff
@@ -67,4 +74,14 @@
       expect(array).to eq(nil)
     end
   end
+
+  describe '#get_ios_app_id_from_archive' do
+    it 'returns identifier ' do
+      expect(CFPropertyList).to receive(:native_types)
+        .and_return(app_path)
+      expect(CFPropertyList).to receive(:native_types)
```
lfkellogg, line 82: Can you just add comments describing what each of these calls is for?

spec/firebase_app_distribution_helper_spec.rb
```diff
@@ -67,4 +74,14 @@
...
+        .and_return(identifier)
+      expect(helper.get_ios_app_id_from_archive("path")).to eq("identifier")
```
lfkellogg, line 84: I would like it if "path" was incorporated into the mocks, so that when reading the tests it is clear that the implementation of `get_ios_app_id_from_archive` actually uses it.

PR #108, opened 2020-07-30
Adding in auth with firebase-tools

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,26 +6,42 @@ module FirebaseAppDistributionAuthClient
...
+          config_path = File.expand_path(".config/configstore/firebase-tools.json", "~")
+        end
         if !google_service_path.nil? && !google_service_path.empty?
           service_account(google_service_path)
         elsif ENV["FIREBASE_TOKEN"]
-          firebase_token
+          firebase_token(ENV["FIREBASE_TOKEN"])
         elsif ENV["GOOGLE_APPLICATION_CREDENTIALS"]
           service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
+        elsif File.exist?(config_path)
```
tagboola, line 20: Consider making `config_path` a private method for readability.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,26 +6,42 @@ module FirebaseAppDistributionAuthClient
...
+          begin
+            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
+            if refresh_token.nil? || refresh_token.empty?
+              UI.crash!(ErrorMessage::MISSING_REFRESH_TOKEN)
```
tagboola, line 24: If we can't find the refresh token in the file I think that's _fine_ and we shouldn't crash. Instead we should fall back to the `UI.crash!(ErrorMessage::MISSING_CREDENTIALS)` since we basically were unable to find any valid auth mechanism.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,26 +6,42 @@ module FirebaseAppDistributionAuthClient
...
+        if ENV["XDG_CONFIG_HOME"]
+          config_path = File.expand_path("configstore/firebase-tools.json", ENV["XDG_CONFIG_HOME"])
+        else
+          config_path = File.expand_path(".config/configstore/firebase-tools.json", "~")
+        end
         if !google_service_path.nil? && !google_service_path.empty?
           service_account(google_service_path)
         elsif ENV["FIREBASE_TOKEN"]
-          firebase_token
+          firebase_token(ENV["FIREBASE_TOKEN"])
```
lfkellogg, line 17: Nice 👍

lfkellogg: Oh, I thought this was for style but it was essentially a refactor to support the new auth method. Still like it!

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,26 +6,42 @@ module FirebaseAppDistributionAuthClient
...
           service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
+        elsif File.exist?(config_path)
+          begin
+            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
+            if refresh_token.nil? || refresh_token.empty?
+              UI.crash!(ErrorMessage::MISSING_REFRESH_TOKEN)
+            else
+              firebase_token(refresh_token)
+            end
+          rescue NoMethodError
```
lfkellogg, line 28: Is this the error that gets raised when the parsed JSON does not have a 'tokens' field? This is kind of a generic error to catch for that case, so I'd prefer checking specifically for that case and like Tunde suggested, falling back to the normal `UI.crash!(ErrorMessage::MISSING_CREDENTIALS)` in the `else`.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,35 +6,42 @@ module FirebaseAppDistributionAuthClient
       TOKEN_CREDENTIAL_URI = "https://oauth2.googleapis.com/token"
 
       def fetch_auth_token(google_service_path)
-        if ENV["XDG_CONFIG_HOME"]
-          config_path = File.expand_path("configstore/firebase-tools.json", ENV["XDG_CONFIG_HOME"])
-        else
-          config_path = File.expand_path(".config/configstore/firebase-tools.json", "~")
-        end
+        config_path = find_firebase_tools
```
lfkellogg: Since this variable is only used in one place, you can remove this and just call the method in-line.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,35 +6,42 @@ module FirebaseAppDistributionAuthClient
...
+           refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
+           unless refresh_token.nil? && refresh_token.empty?
+             refresh_token
+           end
+         rescue NoMethodError
+         end
+        end
+      end
+
+      def find_firebase_tools
```
lfkellogg: I think `firebase_tools_path` would be a more descriptive and Ruby-esque name for this method.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,35 +6,42 @@ module FirebaseAppDistributionAuthClient
...
+        elsif (refresh_token = has_refresh_token(config_path))
+          firebase_token(refresh_token)
         else
           UI.crash!(ErrorMessage::MISSING_CREDENTIALS)
         end
       end
 
       private
 
+      def has_refresh_token(config_path)
```
lfkellogg: I think it would be good to make this method name a little more descriptive. Maybe something like `firebase_tools_refresh_token`.

Also `has_*` implies that it returns a boolean. I think `get_thing` or just `thing` would be better since it returns the refresh token.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,35 +6,42 @@ module FirebaseAppDistributionAuthClient
...
+        if File.exist?(config_path)
+          begin
+           refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
+           unless refresh_token.nil? && refresh_token.empty?
+             refresh_token
+           end
+         rescue NoMethodError
```
lfkellogg, line 37: If you're going to have an empty rescue block, I think a comment is warranted to describe when this case will happen, and why it's OK for it to be empty.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -6,35 +6,42 @@ module FirebaseAppDistributionAuthClient
...
-            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
-            if refresh_token.nil? || refresh_token.empty?
-              UI.crash!(ErrorMessage::MISSING_REFRESH_TOKEN)
-            else
-              firebase_token(refresh_token)
-            end
-          rescue NoMethodError
-            UI.crash!(ErrorMessage::MISSING_REFRESH_TOKEN)
-          end
+        elsif (refresh_token = has_refresh_token(config_path))
```
lfkellogg: I think the assignment here looks great :+1:

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
mannyjimenez0810: I did have get_ before but rubocop yelled for some reason not sure why.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: Oops was looking at an old version

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: I think the current name `refresh_token_from_firebase_tools` is great!

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: Oops was looking at an old version

PR #110, opened 2020-08-04
Auth token refactor

alonsosalasinfante:
Adds firebase cli token parameter auth option, UI informs user of auth method used, refactors error handling and specs

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -21,7 +21,7 @@ module Fastlane
 
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
-        auth_token = fetch_auth_token(params[:service_credentials_file])
+        auth_token = fetch_auth_token(params[:service_credentials_file], params[:firebase_cli_token])
         fad_api_client = Client::FirebaseAppDistributionApiClient.new(auth_token)
         platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
         binary_path = params[:ipa_path] || params[:apk_path]
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
@@ -8,6 +8,7 @@ module Fastlane
       SERVICE_ACCOUNT_PARAM = "service_credentials_file path parameter"
       FIREBASE_TOKEN_ENV = "FIREBASE_TOKEN environment variable"
       FIREBASE_TOKEN_PARAM = "firebase_cli_token parameter"
+      FIREBASE_TOOLS = "Firebase CLI token"
 
       # Returns the auth token for any of the auth methods (Firebase CLI token,
       # Google service account (TODO: firebase-tools). To ensure that a specific
@@ -36,6 +37,9 @@ module Fastlane
         elsif !ENV["GOOGLE_APPLICATION_CREDENTIALS"].nil? && !ENV["GOOGLE_APPLICATION_CREDENTIALS"].empty?
           token = service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
           auth_method = SERVICE_ACCOUNT_ENV
+        elsif (refresh_token = refresh_token_from_firebase_tools)
+          token = firebase_token(refresh_token)
+          auth_method = FIREBASE_TOOLS
         else
           UI.user_error!(ErrorMessage::MISSING_CREDENTIALS)
         end
@@ -44,6 +48,26 @@ module Fastlane
       end
 
       private
+
+      def refresh_token_from_firebase_tools
+        if ENV["XDG_CONFIG_HOME"] && !ENV["XDG_CONFIG_HOME"].empty?
+          config_path = File.expand_path("configstore/firebase-tools.json", ENV["XDG_CONFIG_HOME"])
+        else
+          config_path = File.expand_path(".config/configstore/firebase-tools.json", "~")
+        end
+
+        if File.exist?(config_path)
+          begin
+            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
+            unless refresh_token.nil? || refresh_token.empty?
+              refresh_token
+            end
+          # Rescue is empty to return nil instead of an error
+          # when there is an empty "tokens" field in the firebase-tools json
+          rescue NoMethodError
+          end
+        end
+      end
 
       def firebase_token(refresh_token)
         client = Signet::OAuth2::Client.new(
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
@@ -4,31 +4,57 @@ module Fastlane
   module Auth
     module FirebaseAppDistributionAuthClient
       TOKEN_CREDENTIAL_URI = "https://oauth2.googleapis.com/token"
+      SERVICE_ACCOUNT_ENV = "GOOGLE_APPLICATION_CREDENTIALS environment variable"
+      SERVICE_ACCOUNT_PARAM = "service_credentials_file path parameter"
+      FIREBASE_TOKEN_ENV = "FIREBASE_TOKEN environment variable"
+      FIREBASE_TOKEN_PARAM = "firebase_cli_token parameter"
 
-      def fetch_auth_token(google_service_path)
+      # Returns the auth token for any of the auth methods (Firebase CLI token,
+      # Google service account (TODO: firebase-tools). To ensure that a specific
+      # auth method is used, unset all other auth variables/parameters to nil/empty
+      #
+      # args
+      #   google_service_path - Absolute path to the Google service account file
+      #   firebase_cli_token - Firebase CLI refresh token from login action or
+      #                        CI environment
+      #
+      # env variables
+      #   GOOGLE_APPLICATION_CREDENTIALS - see google_service_path
+      #   FIREBASE_TOKEN - see firebase_cli_token
+      #
+      # Crashes if given invalid or missing credentials
+      def fetch_auth_token(google_service_path, firebase_cli_token)
         if !google_service_path.nil? && !google_service_path.empty?
-          service_account(google_service_path)
-        elsif ENV["FIREBASE_TOKEN"]
-          firebase_token
-        elsif ENV["GOOGLE_APPLICATION_CREDENTIALS"]
-          service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
+          token = service_account(google_service_path)
+          auth_method = SERVICE_ACCOUNT_PARAM
+        elsif !firebase_cli_token.nil? && !firebase_cli_token.empty?
+          token = firebase_token(firebase_cli_token)
+          auth_method = FIREBASE_TOKEN_PARAM
+        elsif !ENV["FIREBASE_TOKEN"].nil? && !ENV["FIREBASE_TOKEN"].empty?
+          token = firebase_token(ENV["FIREBASE_TOKEN"])
+          auth_method = FIREBASE_TOKEN_ENV
+        elsif !ENV["GOOGLE_APPLICATION_CREDENTIALS"].nil? && !ENV["GOOGLE_APPLICATION_CREDENTIALS"].empty?
+          token = service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
+          auth_method = SERVICE_ACCOUNT_ENV
         else
-          UI.crash!(ErrorMessage::MISSING_CREDENTIALS)
+          UI.user_error!(ErrorMessage::MISSING_CREDENTIALS)
         end
+        UI.success("Authenticated with #{auth_method}")
+        token
       end
 
       private
 
-      def firebase_token
+      def firebase_token(refresh_token)
         begin
           client = Signet::OAuth2::Client.new(
             token_credential_uri: TOKEN_CREDENTIAL_URI,
             client_id: Fastlane::Actions::FirebaseAppDistributionLoginAction::CLIENT_ID,
             client_secret: Fastlane::Actions::FirebaseAppDistributionLoginAction::CLIENT_SECRET,
-            refresh_token: ENV["FIREBASE_TOKEN"]
+            refresh_token: refresh_token
           )
         rescue Signet::AuthorizationError
-          UI.crash!(ErrorMessage::REFRESH_TOKEN_ERROR)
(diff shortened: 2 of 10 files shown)


spec/firebase_app_distribution_auth_client_spec.rb
```diff
@@ -29,87 +29,191 @@
...
+        allow(ENV).to receive(:[])
+          .with("GOOGLE_APPLICATION_CREDENTIALS")
+          .and_return(nil)
+        allow(ENV).to receive(:[])
+          .with("FIREBASE_TOKEN")
+          .and_return(nil)
+        allow(ENV).to receive(:[])
+          .with("XDG_CONFIG_HOME")
+          .and_return(nil)
+      end
```
alonsosalasinfante, line 46: Right now specs are doubled because we are testing nil and empty string inputs. Is there a way to make a helper method that can set these allow() calls before the specs run? Having this before(:each) block works but I'm forced to double the specs in order to set all the variables to "" for the empty string tests. It would be easier to have a helper method to set all the environment variables to nil or empty strings and call that within each spec but I am not sure if it's possible to do within a spec file

spec/firebase_app_distribution_auth_client_spec.rb
```diff
@@ -29,87 +29,191 @@
...
     end
 
-    it 'fails if the firebase tools has no refresh_token field' do
-      allow(File).to receive(:read)
-        .and_return(fake_firebase_tools_contents_no_refresh_field)
-      expect(File).to receive(:exist?).and_return(true)
-      expect { auth_helper.fetch_auth_token(nil) }
-        .to raise_error(ErrorMessage::MISSING_CREDENTIALS)
+    describe 'with all other auth variables as empty' do
```
lfkellogg, line 127: Good call to test both the empty and nil cases. I think if you can avoid essentially copying all the tests from above, though, it would be good. Having everything twice makes it harder to maintain these tests. I think there are a couple good alternative options:

- Wrap your tests in a loop that will test one of multiple values, in this case nil and empty ([example](https://github.com/rspec/rspec-core/issues/2663#issuecomment-533634878))
- If you can, select a subset of the tests that will exercise enough of the code to check the nil vs. "" case. For example, would a single test that just ran `auth_client.fetch_auth_token("", "")` and asserted that it raises a `MISSING_CREDENTIALS` be enough to test that the code respects `""` as a nonexistent value?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -4,56 +4,82 @@ module Fastlane
...
         end
+        UI.success("Authenticated with #{auth_method}")
+        token
       end
 
       private
 
       def refresh_token_from_firebase_tools
-        if ENV["XDG_CONFIG_HOME"]
+        if ENV["XDG_CONFIG_HOME"] && !ENV["XDG_CONFIG_HOME"].empty?
```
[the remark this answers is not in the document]
alonsosalasinfante: After some reading some stack overflow it seems like the only two false values are false itself and nil, so there is not really a difference. What do you think is the best way for code style?

spec/firebase_app_distribution_auth_client_spec.rb
alonsosalasinfante: Thanks! I was able to wrap all the specs in a loop, exactly what I was looking for

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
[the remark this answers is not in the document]
tagboola: Up to you. I think the most important thing is that you be consistent throughout the code with whatever method you choose 🙂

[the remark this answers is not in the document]
alonsosalasinfante: Sounds good, I'll go with !ENV["THING"].nil? since we are explicitly checking for nil values

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -4,56 +4,82 @@ module Fastlane
...
       TOKEN_CREDENTIAL_URI = "https://oauth2.googleapis.com/token"
+      SERVICE_ACCOUNT_ENV = "GOOGLE_APPLICATION_CREDENTIALS environment variable"
+      SERVICE_ACCOUNT_PARAM = "service_credentials_file path parameter"
+      FIREBASE_TOKEN_ENV = "FIREBASE_TOKEN environment variable"
+      FIREBASE_TOKEN_PARAM = "firebase_cli_token parameter"
+      FIREBASE_TOOLS = "Firebase CLI token"
 
-      def fetch_auth_token(google_service_path)
+      # Returns the auth token for any of the auth methods (Firebase CLI token,
+      # Google service account (TODO: firebase-tools). To ensure that a specific
```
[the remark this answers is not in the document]
alonsosalasinfante: yup!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -21,7 +21,7 @@ class FirebaseAppDistributionAction < Action
 
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
-        auth_token = fetch_auth_token(params[:service_credentials_file])
+        auth_token = fetch_auth_token(params[:service_credentials_file], params[:firebase_cli_token])
```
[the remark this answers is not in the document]
lfkellogg: Ah, I see what you mean. I see value in both. The benefit to this one is that the Client doesn't have to think about params at all. Don't have a strong opinion, though.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -4,56 +4,87 @@ module Fastlane
...
+            token = service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
+          elsif (refresh_token = refresh_token_from_firebase_tools)
+            auth_method = FIREBASE_TOOLS_DESC
+            token = firebase_token(refresh_token)
+          else
+            UI.user_error!(ErrorMessage::MISSING_CREDENTIALS)
+          end
+        rescue => error
+          unless auth_method.nil? # nil only when no credentials are passed in
+            UI.error("App Distribution encountered an error while authenticating with #{auth_method}")
```
lfkellogg, line 49: The goal of this is to let the customer know which auth method they are using, right? Instead of catching here and printing an extra error, could we just print the auth method when we decide to use it (i.e. a UI.message in each `elsif` block above.

Also catching all errors and wrapping them in `user_error!` seems misleading, since this should only catch unexpected errors, right?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -4,56 +4,87 @@ module Fastlane
...
-             refresh_token
-           end
-         # Rescue is empty to return nil instead of an error
-         # when there is an empty "tokens" field in the firebase-tools json
-         rescue NoMethodError
-         end
+            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
+            refresh_token unless refresh_token.nil? || refresh_token.empty?
+          # Rescue is empty to return nil instead of an error
+          # when there is an empty "tokens" field in the firebase-tools json
```
lfkellogg, line 71: The newline in this comment is in kind of an odd place

spec/firebase_app_distribution_auth_client_spec.rb
```diff
@@ -29,87 +29,101 @@
...
-      .with("XDG_CONFIG_HOME")
-      .and_return(nil)
   end
 
   describe '#fetch_auth_token' do
-    it 'uses service credentials for authorization when the path is passed in' do
-      expect(auth_helper.fetch_auth_token("google_service_path")).to eq("service_fake_auth_token")
-    end
+    describe 'with all other auth variables as nil or empty' do
+      [nil, ""].each do |testval|
```
lfkellogg, line 36: Can you give `testval` a more meaningful name, maybe `nothing_value` or something?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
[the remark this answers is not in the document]
alonsosalasinfante: I think I'll just leave it as is so that only the main action deals with the params

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -4,56 +4,87 @@ module Fastlane
...
         end
+        UI.success("Authenticated with #{auth_method}")
+        token
       end
 
       private
 
       def refresh_token_from_firebase_tools
-        if ENV["XDG_CONFIG_HOME"]
+        if !ENV["XDG_CONFIG_HOME"].nil? && !ENV["XDG_CONFIG_HOME"].empty?
```
tagboola, line 60: The `!` here make this hard to read. We can get rid of them by changing the order of the if else i.e.

```
if ENV["XDG_CONFIG_HOME"].nil? || ENV["XDG_CONFIG_HOME"].empty?
  config_path = File.expand_path(".config/configstore/firebase-tools.json", "~")
else
 config_path = File.expand_path("configstore/firebase-tools.json", ENV["XDG_CONFIG_HOME"]) 
end
```

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
alonsosalasinfante: Good points, I'll use a ui message in each branch instead

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
alonsosalasinfante: good catch!

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -5,55 +5,73 @@ module Auth
...
+          token = firebase_token(firebase_cli_token)
+        elsif !ENV["FIREBASE_TOKEN"].nil? && !ENV["FIREBASE_TOKEN"].empty?
+          UI.message("Authenticating with FIREBASE_TOKEN environment variable")
+          token = firebase_token(ENV["FIREBASE_TOKEN"])
+        elsif !ENV["GOOGLE_APPLICATION_CREDENTIALS"].nil? && !ENV["GOOGLE_APPLICATION_CREDENTIALS"].empty?
+          UI.message("Authenticating with GOOGLE_APPLICATION_CREDENTIALS environment variable")
+          token = service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
         elsif (refresh_token = refresh_token_from_firebase_tools)
-          firebase_token(refresh_token)
+          UI.message("Authenticating with the firebase login:ci command")
```
tagboola, line 36: I think we should be explicit that this is a fallback method because no other authentication method was used. So maybe something like:

"No authentication method specified. Using cached Firebase CLI credentials."

@lfkellogg thoughts?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -5,55 +5,73 @@ module Auth
...
+      #
+      # Crashes if given invalid or missing credentials
+      def fetch_auth_token(google_service_path, firebase_cli_token)
         if !google_service_path.nil? && !google_service_path.empty?
-          service_account(google_service_path)
-        elsif ENV["FIREBASE_TOKEN"]
-          firebase_token(ENV["FIREBASE_TOKEN"])
-        elsif ENV["GOOGLE_APPLICATION_CREDENTIALS"]
-          service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"])
+          UI.message("Authenticating with --service_credentials_file path parameter")
```
tagboola, line 24: Include the path to the service credentials file in the message

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -5,55 +5,73 @@ module Auth
...
+          token = service_account(google_service_path)
+        elsif !firebase_cli_token.nil? && !firebase_cli_token.empty?
+          UI.message("Authenticating with --firebase_cli_token parameter")
+          token = firebase_token(firebase_cli_token)
+        elsif !ENV["FIREBASE_TOKEN"].nil? && !ENV["FIREBASE_TOKEN"].empty?
+          UI.message("Authenticating with FIREBASE_TOKEN environment variable")
+          token = firebase_token(ENV["FIREBASE_TOKEN"])
+        elsif !ENV["GOOGLE_APPLICATION_CREDENTIALS"].nil? && !ENV["GOOGLE_APPLICATION_CREDENTIALS"].empty?
+          UI.message("Authenticating with GOOGLE_APPLICATION_CREDENTIALS environment variable")
```
tagboola, line 33: Include the path to the service credentials file in the message

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -5,55 +5,73 @@ module Auth
...
           begin
-           refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
-           unless refresh_token.nil? && refresh_token.empty?
-             refresh_token
-           end
-         # Rescue is empty to return nil instead of an error
-         # when there is an empty "tokens" field in the firebase-tools json
-         rescue NoMethodError
-         end
+            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
```
tagboola, line 56: I don't think we should address this in this particular CL, but I _think_ we want to catch and swallow any `JSON::ParserError` that `Json.parse` may raise. Also, feels a little scary rescuing a generic `NoMethodError` here. We should probably use some nested if statements. Not as pretty, but less likely to swallow errors that we shouldn't be swallowing.

alonsosalasinfante: good point, I'll leave a comment in the code for this

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: Ooo, yeah I like that!

PR #112, opened 2020-08-05
Adding in metrics

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -91,6 +91,9 @@ def get_upload_token(app_id, binary_path)
       def upload_binary(app_id, binary_path)
         connection.post(binary_upload_url(app_id), File.open(binary_path).read) do |request|
           request.headers["Authorization"] = "Bearer " + @auth_token
+          request.headers["X-APP-DISTRO-API-CLIENT-ID"] = "fastlane"
+          request.headers["X-APP-DISTRO-API-CLIENT-TYPE"] =  Actions.lane_context[Actions::SharedValues::PLATFORM_NAME].to_s
```
mannyjimenez0810, line 95: We did consider making platform a class level variable in action, but the assignment of class level variable during runtime does not seem to be feasible. We had also thought about passing in platform as an instance variable or parameters, but since it is only used once looking for the platform name again didn't seem too expensive.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -68,13 +67,13 @@ def self.details
       end
 
       def self.available_options
-        platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
+        @platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
```
lfkellogg, line 70: Here you are relying on the fact that `available_options` is called before `run`. While this may be the case now, the core Fastlane code that calls these methods could change at any time. Also if someone wanted to modify this class later to use `@platform`, they would have to recognize this and figure out whether or not it would work for their use case. In general, requiring methods to be called in a particular order should be almost always avoided.

A more appropriate pattern here would just be to add a `platform` method that simply returned `Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]`.

If that was an expensive operation, and it was not expected to change, you could memoize it (https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/):
```
def self.platform
  @platform ||= Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
end
```

But I doubt this is expensive. It probably just looks up a value in a hash. So a simple method would work.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
lfkellogg: I thought "client type" was supposed to be the type of client, i.e. gradle, Firebase CLI, etc. Shouldn't "fastlane" go here instead of the platform?

mannyjimenez0810: That is client id. Client type is the android vs ios. I can link you to the developer API docs.

lfkellogg: I see. That conflicts with the Gradle plugin though, AFAICT: https://github.com/FirebasePrivate/firebase-android-buildtools/blob/a4821b09655eebf1593ac32e38c52ad8c9f47afb/firebase-appdistribution/appdistribution-gradle/src/main/java/com/google/firebase/appdistribution/CliRuntimeProperties.java#L23

@tagboola is it possible the docs are wrong: https://docs.google.com/document/d/16wkaDtE0W2dTjfn_dvSGDPRbpTOSjAzhFf_IyARQGwc/edit#heading=h.msoyza6jz9fg

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
mannyjimenez0810: Sounds good I agree with this change.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -23,12 +23,11 @@ def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
         auth_token = fetch_auth_token(params[:service_credentials_file])
         fad_api_client = Client::FirebaseAppDistributionApiClient.new(auth_token)
-        platform = Actions.lane_context[Actions::SharedValues::PLATFORM_NAME]
         binary_path = params[:ipa_path] || params[:apk_path]
 
         if params[:app] # Set app_id if it is specified as a parameter
           app_id = params[:app]
-        elsif platform == :ios
+        elsif @platform == :ios
```
tagboola, line 30: Prefer using  the `platform` method throughout this file instead of using the `@platform` variable directly.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: @lfkellogg I wrote the docs so there is DEFINITELY a possibility that they are wrong 😆 

But in this case setting the client type to ios or android is correct. If you look at code in the Firebase CLI we set the `X-APP-DISTRO-API-CLIENT-TYPE` header  to platform:
https://github.com/firebase/firebase-tools/blob/57d008a5eebc9f091e78c07134ce2625a4d13156/src/appdistribution/client.ts#L62

For some reason in the gradle plugin we set the `X-APP-DISTRO-API-CLIENT-TYPE` to gradle, but I assume the dashboard is smart enough to know that gradle means android.
https://github.com/FirebasePrivate/firebase-android-buildtools/blob/98711268db98522c118585f0ab6728e9ec1a5a38/firebase-appdistribution/appdistribution-gradle/src/main/groovy/com/google/firebase/appdistribution/gradle/UploadDistributionTask.groovy#L117

I think I would prefer this be passed as an instance variable to the api client. Even though it's only used once the client depends on it so I think it's fine to pass it in.

lfkellogg: OK sounds good, thanks guys!

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -3,7 +3,12 @@
   let(:fake_binary_contents) { "Hello World" }
   let(:fake_binary) { double("Binary") }
   let(:headers) { { 'Authorization' => 'Bearer auth_token' } }
-
+  let(:upload_headers) do
```
lfkellogg, line 6: If `upload_headers` is only used inside the `describe '#upload_binary'` block, then you can move this let down there. In general I think it's a good idea to scope these variables as low as possible just to avoid having declaration overload up here at the top of the file.

mannyjimenez0810: That makes sense thank you!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -38,7 +37,7 @@ def self.run(params)
         if app_id.nil?
           UI.crash!(ErrorMessage::MISSING_APP_ID)
         end
-        release_id = fad_api_client.upload(app_id, binary_path)
+        release_id = fad_api_client.upload(app_id, binary_path, platform.to_s)
```
tagboola, line 40: nit: is the `to_s` necessary?

mannyjimenez0810: Yes because if not then it passes in a symbol and when you try to set it in the header it gives you an error. 
Screen Shot 2020-08-06 at 10 12 28 AM

PR #114, opened 2020-08-06
Replacing apk/ipa with the correct binary type

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -1,14 +1,22 @@
 module ErrorMessage
+  def self.binary
```
tagboola, line 2: Consider replacing the constants that use `binary` with methods that take in `binary` as an argument e.g.

```
self.binary_not_found(binary_type)
"Could not find the #{binary_type}. Make sure you set the #{binary_type} path parameter to point to your #{binary_type}"
end
```

PR #115, opened 2020-08-06
Auth creds  error handling

alonsosalasinfante:
Adds method description to get_upload_token

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
@@ -67,11 +67,20 @@ module Fastlane
         UI.success("Release notes have been posted.")
       end
 
+      # Returns the url encoded upload token used for get_upload_status calls:
+      # projects/<project-number>/apps/<app-id>/releases/-/binaries/<binary-hash>
+      #
+      # args
+      #   app_id - Firebase App ID
+      #   binary_path - Absolute path to your app's apk/ipa file
+      #
+      # Throws a user_error if an invalid app id is passed in, the binary file does
+      # not exist, or invalid auth credentials are used (e.g. wrong project permissions)
       def get_upload_token(app_id, binary_path)
         begin
           binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
         rescue Errno::ENOENT
-          UI.crash!("#{ErrorMessage::APK_NOT_FOUND}: #{binary_path}")
+          UI.user_error!("#{ErrorMessage::APK_NOT_FOUND}: #{binary_path}")
         end
 
         begin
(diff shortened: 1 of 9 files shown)


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -85,11 +94,13 @@ def get_upload_token(app_id, binary_path)
             request.headers["Authorization"] = "Bearer " + @auth_token
           end
         rescue Faraday::ResourceNotFound
-          UI.crash!("#{ErrorMessage::INVALID_APP_ID}: #{app_id}")
+          UI.user_error!("#{ErrorMessage::INVALID_APP_ID}: #{app_id}")
+        rescue Faraday::ForbiddenError
```
tagboola, line 98: Technically any call can raise a `Faraday::ForbiddenError` error, so we should either rescue it in every call or just let it fail loudly. I'm okay with letting it fail loudly i.e. removing the `rescue Faraday::ForbiddenError`.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -85,11 +94,13 @@ def get_upload_token(app_id, binary_path)
...
+          UI.user_error!("#{ErrorMessage::INVALID_CREDENTIALS}: #{app_id}")
         end
         contact_email = response.body[:contactEmail]
         if contact_email.nil? || contact_email.strip.empty?
-          UI.crash!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
+          UI.user_error!(ErrorMessage::GET_APP_NO_CONTACT_EMAIL_ERROR)
```
tagboola, line 103: Nice!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
alonsosalasinfante: The reason I wanted to catch this error is because using the wrong service account credentials will allow the user to successfully authenticate but throw a 403 when trying to get the app token, so I wanted to make the error message clearer to reflect that.

tagboola: Gotcha. That makes sense. The reason I think I'm okay with not catching this and just failing loudly is because a generic 403 can be returned for a lot of different reasons, not just when a developer is using the wrong service account credentials. For example, if you use a refresh token for a user that doesn't have access to the app that will return a 403. Or if a developer is using the correct service account credentials but doesn't have the right permissions a 403 will be returned (right now this isn't possible i.e. anyone who has access can at least check the upload status, but in the future you can imagine a world where there is an "uploader role" that only allows uploading binaries and "manager" role which allows managing testers).

PR #116, opened 2020-08-06
Catching type error

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -76,7 +76,7 @@ def post_notes(app_id, release_id, release_notes)
       def get_upload_token(app_id, binary_path)
         begin
           binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
-        rescue Errno::ENOENT
+        rescue Errno::ENOENT, TypeError
```
tagboola, line 79: Can you explain why this is needed?

mannyjimenez0810: when the apk or ipa isnt passed in and its nil it tries to convert the binary path into a string. It cannot convert nil to string and so we get this error. This is just catching that error and giving the user a better error message
Screen Shot 2020-08-06 at 2 18 34 PM

lfkellogg: It's pretty hard to make that connection just looking at the code. A TypeError is a pretty vague. I think it would be more readable/explicit if we were to check the `binary_path` explicitly to make sure it's not nil.

In fact I think the same thing goes for the Errno::ENOENT. Could we instead just have a check that crashes if either binary_path is nil or the file does not exist?

mannyjimenez0810: That makes sense. I have included a check that does check for the file and if its nil. If the user gives in a path to a file that doesn't exist as a parameter for apk or the ipa then it will fail in the options. I added the file exist check as an extra precaution although it may be unnecessary.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -74,9 +74,9 @@ def post_notes(app_id, release_id, release_notes)
       end
 
       def get_upload_token(app_id, binary_path)
-        begin
+        if !binary_path.nil? && File.exist?(binary_path)
           binary_hash = Digest::SHA256.hexdigest(File.open(binary_path).read)
-        rescue Errno::ENOENT
+        else
```
lfkellogg, line 79: In this case, where there is a specific error case to check before moving on, I would avoid the `else` altogether:

```
if (error case)
  crash!
end

binary_hash = thing
```

In this case it also takes the assignment out of the conditional, which is nice.

PR #118, opened 2020-08-10
Login ui refactor

alonsosalasinfante:
Refactors ui messages

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
@@ -22,7 +22,7 @@ module Fastlane
         code = UI.input("Enter the resulting code here: ")
         credentials = authorizer.get_credentials_from_code(code: code, base_url: OOB_URI)
 
-        UI.success("Set the following refresh token as the FIREBASE_TOKEN environment variable by running the following command:")
+        UI.success("Set the refresh token as the FIREBASE_TOKEN environment variable by running the following command:")
         UI.success("\"export FIREBASE_TOKEN=#{credentials.refresh_token}\"")
       rescue Signet::AuthorizationError
         UI.error("The code you entered is invalid. Copy and paste the code and try again.")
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -16,20 +16,19 @@ def self.run(params)
         authorizer = Google::Auth::UserAuthorizer.new(client_id, SCOPE, nil)
         url = authorizer.get_authorization_url(base_url: OOB_URI)
 
-        UI.message("Please open the following address in your browser:")
+        UI.message("Open the following address in your browser and sign in with the Google account on your Firebase project:")
```
tagboola, line 19: I get what your trying to convey here, but the part that says “on your Firebase project” sounds weird. Consider removing it. I think it’s _okay_ to assume that the google account they use will be the same one that has access to their firebase project.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -16,20 +16,19 @@ def self.run(params)
...
         UI.message(url)
         UI.message("")
         code = UI.input("Enter the resulting code here: ")
         credentials = authorizer.get_credentials_from_code(code: code, base_url: OOB_URI)
 
-        UI.message("Refresh Token: #{credentials.refresh_token}")
-        UI.message("")
-        UI.message("Set the refresh token as a FIREBASE_TOKEN environment variable")
+        UI.success("Set the refresh token as the FIREBASE_TOKEN environment variable by running the following command:")
+        UI.success("\"export FIREBASE_TOKEN=#{credentials.refresh_token}\"")
```
tagboola, line 26: I think I would leave this the way it was. There are other ways to set an environmental variable, e.g you can add the `export` statement to your `bashrc` file, so I think we can leave it up to the developer to decide how to do it.

alonsosalasinfante: That makes sense, I'll keep the message->success change to make the refresh token more visible

PR #121, opened 2020-08-18
Remove dependency on Firebase CLI being installed

lfkellogg:
Because it's set as required, it will always use the default value and try to verify that the tool is installed. This change makes it optional and removes the validation since we never use it anyway.

We should follow this up with some more cleanup to remove the dead code surrounding this as well as to add a message saying that the firebase_cli_path parameter is deprecated, if it is set by the user.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -13,7 +13,6 @@ require_relative '../helper/firebase_app_distribution_auth_client'
 module Fastlane
   module Actions
     class FirebaseAppDistributionAction < Action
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -108,20 +108,8 @@ def self.available_options
           FastlaneCore::ConfigItem.new(key: :firebase_cli_path,
                                        env_name: "FIREBASEAPPDISTRO_FIREBASE_CLI_PATH",
                                        description: "The absolute path of the firebase cli command",
-                                       default_value: DEFAULT_FIREBASE_CLI_PATH,
```
tagboola, line 111: Remove unused `DEFAULT_FIREBASE_CLI_PATH`

PR #127, opened 2020-08-19
Removes this APK has not been uploaded before

alonsosalasinfante:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -140,10 +140,7 @@ def upload_binary(app_id, binary_path, platform)
       def upload(app_id, binary_path, platform)
         upload_token = get_upload_token(app_id, binary_path)
         upload_status_response = get_upload_status(app_id, upload_token)
-        if upload_status_response.success? || upload_status_response.already_uploaded?
-          UI.success("This #{@binary_type} has been uploaded before. Skipping upload step.")
```
lfkellogg, line 144: I actually think this message is OK to leave in. If we can detect that it's already uploaded I think it's good to tell the user. I was more concerned about the other message, since it might be wrong.

PR #137, opened 2020-08-31
Use binary path to determine path, if none set

lfkellogg:
Previously we relied only on the lane context to get the platform. This was causing a surprising number of `unknown` platforms to show up in our metrics. In the Firebase CLI we use the binary file type exclusively to determine the platform. Here we'll use it as a backup if the lane context isn't set.

This also:
- Fixes some failing tests (for me locally, anyway) due to not calling through when mocking the `File` object
- Removes the unnecessary memoization of the lane platform, since it's just a hash lookup

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -21,8 +21,9 @@ module Fastlane
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk
         auth_token = fetch_auth_token(params[:service_credentials_file], params[:firebase_cli_token])
-        fad_api_client = Client::FirebaseAppDistributionApiClient.new(auth_token, platform)
         binary_path = params[:ipa_path] || params[:apk_path]
+        platform = lane_platform || platform_from_path(binary_path)
+        fad_api_client = Client::FirebaseAppDistributionApiClient.new(auth_token, platform)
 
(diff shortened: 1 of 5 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -65,16 +66,26 @@ def self.details
...
+        when 'ipa'
+          :ios
+        when 'apk'
+          :android
+        end
       end
 
       def self.available_options
-        if platform == :ios || platform.nil?
+        if lane_platform == :ios || lane_platform.nil?
```
tagboola, line 84: Should we fall back `platform_from_path` here if `lane_platform` is nil as well?

lfkellogg: Unfortunately there's kind of a chicken-or-the-egg situation here, since `platform_from_path` depends on knowing the binary path, which is what we're trying to actually set here.

Luckily this block only comes into play if the user has not set an explicit path to the binary, and they also don't have a build block in the same lane, in which case we get the binary path from the lane context. So this should be an edge case.

If none of those are set, and neither is the platform, ideally we'd probably just look for both IPAs and APKs and use whichever one we find to determine the platform, but I think this is OK for now since again it should be kind of an edge case.

tagboola: Gotcha that makes sense. Thanks for the detailed explanation!

Issue #146, opened 2020-09-15
Invalid APK file after upgrading to v0.2.0


tad3j:
I've updated fastlane-plugin-firebase_app_distribution to v0.2.0 and I started getting error bellow when deploying to firebase:

```
App Distribution halted because it had a problem uploading the APK: There's been an error processing your distribution. Ensure you are uploading a valid IPA or APK and try again.
Additional information about this error:
Invalid APK file.
```
If I try to upload the APK file manually via their WEB service file gets uploaded without any issues. Also downgrading to v0.1.4 solves this issue and again lets fastlane upload apk file to fireabse app distribution server. 

Did I miss any breaking changes or could this be an issue with v0.2?

cvaghela:
That is weird, if it works in the console and the previous plugin my guess is that we are somehow uploading the wrong file. Are you passing the `apk_path` or do you let the plugin find the file? Can you share your `Fastfile`?

tad3j:
Yes, I am passing the `apk_path` and I haven't changed that since v0.1.4, so I think it should be pointing to the right file.

Maybe I should also mention that I use monorepo (in case this may be the issue?)...that's why I also pass `firebase_cli_path`, else plugin is looking for cli in wrong folder. Also, I'm running this on Windows.

Here are the relevant parts of my Fastfile:

```
default_platform(:android)

platform :android do
  desc "Yarn Install Dependencies"
  lane :yarn_install do |options|
    yarn(
      command: 'install',
      package_path: '../../..//package.json',
      options: '--ignore-optional --frozen-lockfile'
     )
  end

  desc "Build APK"
  lane :build_apk do |options|
    gradle(task: "clean")
    gradle(
      task: "assemble",
      build_type: options[:release] ? 'Release' : 'Releasestaging',
      print_command: false,
      properties: {
        "android.injected.signing.store.file" => "project.keystore",
        "android.injected.signing.store.password" => options[:storePass],
        "android.injected.signing.key.alias" => "project",
        "android.injected.signing.key.password" => options[:keyPass],
      }
    )
  end

  desc "Distribute (Firebase App Distribution)"
  lane :staging do |options|
    yarn_install()
    build_apk(
      release: false,
      storePass: options[:storePass],
      keyPass: options[:keyPass]
    )
    firebase_app_distribution(
      app: options[:firebaseApp],
      groups: "group-1",
      firebase_cli_path: "../../../node_modules/.bin/firebase",
[...]

PR #148, opened 2020-09-16
Fix binary upload on Windows

lfkellogg:
Should fix https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/146

I was able to reproduce the issue on Windows, and this fixes it. The issue is probably the EOL conversion:

```
"b"  |  Binary file mode (may appear with
     |  any of the key letters listed above).
     |  Suppresses EOL <-> CRLF conversion on Windows. And
     |  sets external encoding to ASCII-8BIT unless explicitly
     |  specified.
```

Tested by uploading new iOS and Android distributions and verifying that the uploads succeed and the apps function correctly.


tagboola (approved):
LGTM to me once you update the failing specs! 🙂

Issue #151, opened 2020-09-17
APK is not zip aligned


FrancisLeigh:
I'm having issues with my React-Native project where i can't seem to build my app to firebase distribution after abstracting my lanes out from their android/ & ios/ folders to be consumed by a "master" fastlane in the root. 

When getting to the `firebase_app_distribution` step, it errors out with 
```bash
+------+------------------------------+-------------+
|                 fastlane summary                  |
+------+------------------------------+-------------+
| Step | Action                       | Time (in s) |
+------+------------------------------+-------------+
| 1    | ensure_git_status_clean      | 0           |
| 2    | npm run test                 | 2           |
| 3    | Switch to android_stage lane | 0           |
| 4    | android_appicon              | 1           |
| 5    | increment_version_name       | 0           |
| 6    | increment_version_code       | 0           |
| 7    | get_version_name             | 0           |
| 8    | get_version_code             | 0           |
| 9    | add_badge                    | 0           |
| 10   | assembleRelease              | 17          |
| 💥   | firebase_app_distribution    | 70          |
+------+------------------------------+-------------+

[13:45:33]: fastlane finished with errors

[!] App Distribution halted because it had a problem uploading the IPA: There's been an error processing your app because of an archiving issue. Use zipalign in your build process and try again. Learn more at https://developer.android.com/studio/command-line/zipalign If this problem continues, contact Firebase support.
Additional information about this error:
APK is not zip aligned.
```

### `root/fastlane/Fastfile`
```ruby
import "../android/fastlane/Fastfile"
import "../ios/fastlane/Fastfile"

desc "Deploy IOS & Android staging build to Firebase App Distribution"
desc "```
@param   : bump_type
@options : major|minor|patch
@dafult  : minor
```"
desc "```
$ fastlane stage bump_type:patch
```"
lane :stage do |options|
    ensure_git_status_clean
    sh "npm run test"
    UI.important("
        No option bump_type provided, bump_type will default to minor...
    ") unless options[:bump_type]

    answer = UI.select("Would you like to add a Changelog?", ["Yes", "No"])
    changelog = ask("Type Changelog message:") if answer == "Yes"

    if changelog
        android_stage(bump_type: options[:bump_type], changelog: changelog)
        ios_stage(bump_type: options[:bump_type], changelog: changelog)
    else
        android_stage(bump_type: options[:bump_type])
        ios_stage(bump_type: options[:bump_type])
    end

    unless options[:no_commit]
[...]

PR #152, opened 2020-09-17
Get binary path from platform, and platform from app ID

lfkellogg:
Previously if no platform is set, we would get the binary path using a combination of the parameters passed in, and searching for an IPA or APK. Then we would determine the platform from the binary path.

This leads to bugs like https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/151, because if for example we find an IPA in the root directory, we will upload that binary regardless of whether or not apk_path is set.

It will be more reliable to get the platform from the app ID, and they use that to determine which path to use.

### DIFF
--- a/spec/firebase_app_distribution_action_spec.rb
+++ b/spec/firebase_app_distribution_action_spec.rb
@@ -1,20 +1,38 @@
 describe Fastlane::Actions::FirebaseAppDistributionAction do
   let(:action) { Fastlane::Actions::FirebaseAppDistributionAction }
-  describe '#platform_from_path' do
-    it 'returns :android when the binary is an APK' do
-      expect(action.platform_from_path('/path/to/your_binary.apk')).to eq(:android)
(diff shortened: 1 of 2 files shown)


spec/firebase_app_distribution_action_spec.rb
```diff
@@ -1,20 +1,38 @@
...
+
+    it 'returns the apk_path for an Android app' do
+      expect(action.binary_path_from_platform(:android, '/ipa/path', '/apk/path')).to eq('/apk/path')
+    end
+
+    it 'returns the ipa_path by default when there is no platform' do
+      expect(action.binary_path_from_platform(nil, '/ipa/path', '/apk/path')).to eq('/ipa/path')
+    end
+
+    it 'falls back on the apk_path when there is no platform' do
```
tagboola, line 30: nit: "falls back on the apk_path when there is no platform or ipa_path"

tagboola (approved):

PR #157, opened 2020-09-22
Add tests around getting the app ID

lfkellogg:


spec/firebase_app_distribution_action_spec.rb
```diff
@@ -59,4 +59,46 @@
...
+
+    it 'raises if the app parameter is not set and there is no archive path' do
+      allow(action).to receive(:xcode_archive_path).and_return(nil)
+
+      params = {}
+      expect { action.app_id_from_params(params) }
+        .to raise_error(ErrorMessage::MISSING_APP_ID)
+    end
+
+    it 'returns the app id from the plist if the archive path is set' do
```
tagboola, line 81: Do we already have an existing test that for the `get_ios_app_id_from_archive_plist` method? If not may be worth adding one to prevent future regressions around parsing the plist file.

tagboola (approved):

spec/firebase_app_distribution_action_spec.rb
lfkellogg: We do, it's over in the helper spec, since that method is in the helper (although it probably doesn't need to be): https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/master/spec/firebase_app_distribution_helper_spec.rb#L75

Issue #159, opened 2020-09-29
The server responded with status 400


dummyco:
Hello!
We've started getting 400 errors right after 0.2.2 update:

```
[20:07:57]: ---------------------------------------
[20:07:57]: --- Step: firebase_app_distribution ---
[20:07:57]: ---------------------------------------
[20:07:57]: Authenticating with GOOGLE_APPLICATION_CREDENTIALS environment variable: /home/circleci/google-service-account.json
[20:07:57]: 🔐 Authenticated successfully.
[20:07:57]: ⌛ Uploading the APK.
[20:08:03]: ✅ Uploaded the APK.
+------------------+---------+
|        Lane Context        |
+------------------+---------+
| DEFAULT_PLATFORM | android |
| PLATFORM_NAME    |         |
| LANE_NAME        | beta    |
+------------------+---------+
[20:08:03]: the server responded with status 400

+------+----------------------------+-------------+
|                fastlane summary                 |
+------+----------------------------+-------------+
| Step | Action                     | Time (in s) |
+------+----------------------------+-------------+
| 1    | opt_out_usage              | 0           |
| 2    | Verifying fastlane version | 0           |
| 3    | default_platform           | 0           |
| 💥   | firebase_app_distribution  | 6           |
+------+----------------------------+-------------+

[20:08:03]: fastlane finished with errors

Looking for related GitHub issues on fastlane/fastlane...

➡️  Distribute to hockeyapp suddenly failing
    https://github.com/fastlane/fastlane/issues/400 [closed] 8 💬
    22 Sep 2016

➡️  Uploading assets with a GitHub release fails in `set_github_release`
    https://github.com/fastlane/fastlane/issues/9922 [closed] 5 💬
    18 Nov 2017

➡️  Deliver: Segmentation fault after completing
    https://github.com/fastlane/fastlane/issues/10479 [closed] 16 💬
    10 May 2018

and 1 more at: https://github.com/fastlane/fastlane/search?q=the%20server%20responded%20with%20status%20400&type=Issues&utf8=✓

bundler: failed to load command: fastlane (/home/circleci/.rubies/ruby-2.6.1/bin/fastlane)
Faraday::BadRequestError: [!] the server responded with status 400
  /home/circleci/.rubies/ruby-2.6.1/lib/ruby/gems/2.6.0/gems/faraday-1.0.1/lib/faraday/response/raise_error.rb:16:in `on_complete'
[...]

PR #160, opened 2020-09-29
fix: specify content-type headers when making json post requests

brownoxford:
Fixes issue #159


brownoxford:
> @googlebot I signed it!

joshdholtz (approved):
Thanks for this fix! I just came to the same solution but I had to go thread setting up a new app on Firebase to test first 🤦‍♂️ I pulled down your changes and I can confirm that they do fix this!

I’ll get the version bumped and new version released right away! Thank you again ❤️

ravitripathi:
Can confirm this fixes https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/159. Our builds are back to normal. Thanks!

PR #162, opened 2020-09-29
Set Content-Type to application/octet-stream for uploads

tagboola:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -39,8 +42,8 @@ def enable_access(app_id, release_id, emails, group_ids)
         payload = { emails: emails, groupIds: group_ids }
         begin
           connection.post(enable_access_url(app_id, release_id), payload.to_json) do |request|
-            request.headers["Authorization"] = "Bearer " + @auth_token
-            request.headers["Content-Type"] = "application/json"
+            request.headers[AUTHORIZATION] = "Bearer " + @auth_token
+            request.headers[CONTENT_TYPE] = "application/json"
```
lfkellogg, line 46: I like the use of constants, should we use them for the MIME types themselves as well?

lfkellogg (approved):

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: Done!

lfkellogg (commented):
Looks great, thanks!

PR #166, opened 2020-10-09
Show error message when posting release notes fails

tagboola:
Parse error response of call to create release notes and display error message.

Screen Shot 2020-10-09 at 9 32 46 AM


lfkellogg (approved):
Nice!

PR #167, opened 2020-10-09
Add support for debug logging

tagboola:
Before `v.0.2.0` we passed the `debug` parameter to the Firebase CLI and it printed verbose output. Now that there is no longer a dependency on the Firebase CLI , this parameter wasn't being used. This change adds support for debug logging the request and the response body.

Screen Shot 2020-10-09 at 9 47 29 AM


lfkellogg (approved):
🙌

PR #175, opened 2020-11-03
Removed the duplicate tester emails to prevent error 500

return-main:
I just stumbled across an edge case in the enable_access step.
Firebase returns an error 500 if there are duplicate email addresses, so I removed the duplicates.


tagboola:
@return-main the 500 error you were seeing for duplicate emails was actually a result of a bug on our backend 😢. We have a fix that will be deployed tomorrow, so I'm going to close this PR since it will no longer be needed. Thanks for bringing this up tho!

PR #183, opened 2020-12-15
Only print invalid app message on 404's returned from get_app call

tagboola:
We are incorrectly assuming that 404's returned from calls to other methods mean an invalid app id was used. This is only true for calls to `get_app`.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
@@ -37,7 +37,7 @@ module Fastlane
       #   emails - String array of app testers' email addresses
       #   group_ids - String array of Firebase tester group IDs
       #
-      # Throws a user_error if app_id, emails, or group_ids are invalid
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -37,7 +37,7 @@ def initialize(auth_token, platform, debug = false)
       #   emails - String array of app testers' email addresses
       #   group_ids - String array of Firebase tester group IDs
       #
-      # Throws a user_error if app_id, emails, or group_ids are invalid
+      # Throws a user_error emails, or group_ids are invalid
```
lfkellogg, line 40: "Throws a user_error *if* emails"

lfkellogg (approved):
Looks good!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: Good catch! Done!

PR #187, opened 2021-02-25
Add clarifying comment about OAuth client secret

lfkellogg:
Similar to the corresponding code in the Firebase CLI: https://github.com/firebase/firebase-tools/blob/2dc7216a498dee2ca7e2acc33d6ba16d5647e27f/src/api.js#L87


tagboola (approved):
Nice!

PR #201, opened 2021-04-26
Exit with a non-zero status code if upload does not finish

lfkellogg:
If after 2 minutes the status shows the binary is still being processed, then exit with an error. Otherwise we are hiding the fact that that distribution may not be uploaded successfully.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
@@ -127,10 +127,10 @@ module Fastlane
       #   app_id - Firebase app ID
       #   binary_path - Absolute path to your app's aab/apk/ipa file
       #
-      # Returns the release_id on a successful release, otherwise returns nil.
+      # Returns the release_id of the uploaded release.
       #
-      # Throws a UI error if the number of polling retries exceeds MAX_POLLING_RETRIES
-      # Crashes if not able to upload the binary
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -159,8 +159,7 @@ def upload(project_number, app_id, binary_path, platform)
             end
           end
           unless upload_status_response.success?
-            UI.error("It took longer than expected to process your #{binary_type}, please try again.")
```
tagboola, line 162: Curious: why crash here versus crashing in the action?

tagboola (approved):
🚀

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
lfkellogg: Because it lets us crash loudly with a relevant message. If we crash when we get a nil release_id back in the action, it feels weird having a specific error message about the cause of the nil, so we'd probably want to log something here anyway.

This reminds me though, we don't want to expect nil to be returned anymore, since it only happened in this case. Removed that and updated the documentation on the method.

tagboola: 👍 👏

Issue #215, opened 2021-07-09
App Distribution could not generate credentials from the service credentials file specified.


nyarian:
### Issue Description

I have the plugin added to the configuration and `service_credentials_file`, `app`, `groups` and `release_notes` parameters specified for the action inside the lane. All the parameters are valid and the service account has `Firebase App Distribution Admin` role given. Nevertheless, the action fails with this error message: "[!] App Distribution could not generate credentials from the service credentials file specified. Service Account Path: /path/to/app/android/app/firebase_app_distribution_account_key.json". AppFile has a valid application id inside the `package_name` method. The same configuration applied to the Gradle's Firebase App Distribution Plugin works and the app was submitted successfully via plugin's task (appDistributionUploadRelease).

##### Configuration
Here's the complete lane's definition including the distribution action:

```ruby
  desc "Distributes the staging app via Firebase App Distribution"
  lane :distribute_firebase do
    build_android_app(
      task: "assemble",
      build_type: "Release"
    )

    firebase_app_distribution(
      service_credentials_file: "#{Dir.pwd.chomp("/fastlane")}/app/firebase_app_distribution_account_key.json",
      app: "app-id-copied-from-the-firebase-app-page",
      groups: "staging",
      release_notes: "Stub"
    )
  end
```

##### Complete verbose output

> user:~/path/to/app/android$ bundle exec fastlane android distribute_firebase
> [✔] 🚀 
> +-------------------------------------------+---------+-------------------------------------+
> |                                       Used plugins                                        |
> +-------------------------------------------+---------+-------------------------------------+
> | Plugin                                    | Version | Action                              |
> +-------------------------------------------+---------+-------------------------------------+
> | fastlane-plugin-firebase_app_distribution | 0.2.9   | firebase_app_distribution,          |
> |                                           |         | firebase_app_distribution_login,    |
> |                                           |         | firebase_app_distribution_get_udids |
> +-------------------------------------------+---------+-------------------------------------+
> 
> [03:11:02]: ------------------------------
> [03:11:02]: --- Step: default_platform ---
> [03:11:02]: ------------------------------
> [03:11:02]: Driving the lane 'android distribute_firebase' 🚀
> [03:11:02]: -----------------------------
> [03:11:02]: --- Step: assembleRelease ---
> [03:11:02]: -----------------------------
> [03:11:02]: $ /path/to/app/android/gradlew assembleRelease -p .
> [03:11:03]: ▸ > ...
[...]

PR #217, opened 2021-07-09
Print details of service account auth errors

lfkellogg:
This will help troubleshoot in cases like https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/215.

Before:
```
[!] App Distribution could not generate credentials from the service credentials file specified. Service Account Path: <path>
```

After:
```
[!] App Distribution could not generate credentials from the service credentials file specified. Service Account Path: <path>

Details:
Authorization failed.  Server message:
{"error":"invalid_grant","error_description":"Invalid grant: account not found"}
Response status: 400
```

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
@@ -13,28 +13,29 @@ module Fastlane
       #   google_service_path - Absolute path to the Google service account file
       #   firebase_cli_token - Firebase CLI refresh token from login action or
       #                        CI environment
+      #   debug - Whether to enable debug-level logging
       #
       # env variables
       #   GOOGLE_APPLICATION_CREDENTIALS - see google_service_path
       #   FIREBASE_TOKEN - see firebase_cli_token
       #
       # Crashes if given invalid or missing credentials
-      def fetch_auth_token(google_service_path, firebase_cli_token)
+      def fetch_auth_token(google_service_path, firebase_cli_token, debug = false)
         if !google_service_path.nil? && !google_service_path.empty?
           UI.message("Authenticating with --service_credentials_file path parameter: #{google_service_path}")
-          token = service_account(google_service_path)
+          token = service_account(google_service_path, debug)
         elsif !firebase_cli_token.nil? && !firebase_cli_token.empty?
           UI.message("Authenticating with --firebase_cli_token parameter")
-          token = firebase_token(firebase_cli_token)
+          token = firebase_token(firebase_cli_token, debug)
         elsif !ENV["FIREBASE_TOKEN"].nil? && !ENV["FIREBASE_TOKEN"].empty?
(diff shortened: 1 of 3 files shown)


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -83,8 +83,11 @@ def service_account(google_service_path)
         service_account_credentials.fetch_access_token!["access_token"]
       rescue Errno::ENOENT
         UI.user_error!("#{ErrorMessage::SERVICE_CREDENTIALS_NOT_FOUND}: #{google_service_path}")
-      rescue Signet::AuthorizationError
-        UI.user_error!("#{ErrorMessage::SERVICE_CREDENTIALS_ERROR}: #{google_service_path}")
+      rescue Signet::AuthorizationError => error
+        UI.user_error!(
```
tagboola, line 87: Should we only display this when `debug` is set to `true`?

lfkellogg: If this call fails I suspect it's always helpful to have more info. If we were showing details about successful requests here too, I think that would make sense to show in debug mode only. FWIW I looked for a "debug" option in the Google auth library but didn't find anything like that.

tagboola: Gotcha. Should we consider showing messages if any of the steps fail? Or is there a particular reason it will be useful to show this message even if the debug flag isn't set?

tagboola (approved):

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: In this case the error on its own is not helpful since there's so little information. Same with the `firebase_token` method (I added this to that one as well). But maybe since we're just dumping the response it feels more like debug level info? I'd still probably lean towards giving too much info vs too little in the case of an error though.

tagboola: Gotcha. I agree probably want to bias showing too much info vs. too little. Are we sure the message won't have any sensitive information? There's an understanding that when the debug flag is set we may show sensitive information, but that's not necessarily the case when it is not set. If we think this won't display any sensitive information LGTM 🙂

tagboola (approved):

tonybaroneee (approved):

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: That's a good point. It probably won't, but seems better to just require the debug flag. I think that will be most folks first instinct anyway when they're seeing an error they don't understand.

Updated.

tagboola (approved):
🚢

PR #219, opened 2021-08-11
Switch endpoints over to use new v1 public API

tonybaroneee:


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -29,32 +29,34 @@ def self.run(params)
         auth_token = fetch_auth_token(params[:service_credentials_file], params[:firebase_cli_token])
         fad_api_client = Client::FirebaseAppDistributionApiClient.new(auth_token, params[:debug])
 
-        # If binary is an AAB get FULL view of app which includes the aab_state
-        app_view = binary_type == :AAB ? 'FULL' : 'BASIC'
-        app = fad_api_client.get_app(app_id, app_view)
-        validate_app!(app, binary_type)
+        app = fad_api_client.get_app(app_id)
```
tagboola, line 32: Ideally we would like to no longer depend on this `v1alpha` endpoint. Consider just parsing the project number from the app ID or passing a "-" as the project number (not sure if we currently support this on the backend, but it's technically a valid use case https://google.aip.dev/159). I _think_ we could also use the `projects.androidApps` or `projects.iosApps` endpoints to get the project number.

https://firebase.google.com/docs/projects/api/reference/rest

tonybaroneee: We need it to validate that a contact email exists.

tagboola: Yeah true but the the hope is that we could accomplish this by throwing an error on the backend instead of having to manually check the contact email address. The developer doesn't really need to know what the contact email is for the app and there is no way to update it using the API so I think this way would be preferable.

tonybaroneee: We use the project number for IAM stuff / backend permission checking, so I'm afraid passing `-` would not work. Could we just parse out the project number from the app id, or is that considered not kosher? I think calling those other endpoints would sort of defeat the purpose, since ideally we don't have to make any extra call here.

kaibolay: I parse the project number and ist atrocious: https://github.com/firebase/firebase-tools/blob/kb-use-appdistro-v1-public-api/src/appdistribution/client.ts#L71

I think we should seriously consider supporting `-` on the server (since we can extract the project number from the `appId`)... But until we do we might need to do the parsing on the client 🤮

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
```diff
@@ -0,0 +1,42 @@
+class AabInfo
+  # AAB states
+  class AabState
+    UNSPECIFIED = "AAB_STATE_UNSPECIFIED"
```
jladieu, line 4: Small nit, since these strings don't have any variable interpolation, you can use single quotes instead of doubles.  This style guideline might be implemented haphazardly in the surrounding code, but I'll typically make this adjustment when changing or moving lines to have the code converge on this convention.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
```diff
@@ -0,0 +1,42 @@
...
+    PLAY_ACCOUNT_NOT_LINKED = "PLAY_ACCOUNT_NOT_LINKED"
+    NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT = "NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT"
+    APP_NOT_PUBLISHED = "APP_NOT_PUBLISHED"
+    PLAY_IAS_TERMS_NOT_ACCEPTED = "PLAY_IAS_TERMS_NOT_ACCEPTED"
+    ACTIVE = "ACTIVE"
+    UNAVAILABLE = "AAB_STATE_UNAVAILABLE"
```
jladieu, line 10: This looks like an enum that might be present in backend code, eg: http://google3/java/com/google/firebase/console/web/modules/appdistro/ng2/view/onboarding_state.ts?l=52-60&rcl=389870755 

It'd probably be good to crossref this fastlane definition from that location with a suggestion to keep these enums in sync.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
```diff
@@ -0,0 +1,42 @@
...
+  def initialize(response)
+    @response = response || {}
+  end
+
+  def integration_state
+    @response[:integrationState]
+  end
+
+  def test_certificate
+    @response[:testCertificate]
```
jladieu, line 22: Small DRY nit - you could implement this as:
```
def test_certificate
  @response[:testCertificate] || {}
end
```

And then dispense with the ternary `test_certificate ? test_certificate[:<key>] : nil` in favor of just`test_certificate[:<key>]` in the other three methods.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
jladieu: I have concerns with parsing the projectNumber from the appId given the format of that id isn't under our control and has changed in the past, unless we have a solid "go for it" from the providers of that ID.

Maybe we can provide an offramp for customers to provide it explicitly as a param, and fall back to parsing it from appId (with a customer-visible deprecation warning or something encouraging them to specify it explicitly), similar to how we provide a couple ways to figure out [platform](https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/pull/219/files#diff-1f50a993e035d5ff62885ac3fddac664536c4ca69bb273ea666d890d3b2b5621L22)

jladieu: Another note: are we missing a public API endpoint that lets us check for the presence of a contact email that necessitates our use of the alpha endpoint?

tonybaroneee: Nope, that’s more of a legacy concept/artifact. All onboarded apps will have a contact email set, and if they aren’t onboarded, our endpoints will return a 404 since they won’t be in spanner. I think we should just parse out the project number from the app ID, since if the app ID format ever changes, we have bigger fish to fry anyways.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
tonybaroneee: Agreed, double quoted made me nauseous but was following the surrounding code :) will update.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def sha1_certificate_hash
+    test_certificate ? test_certificate[:hashSha1] : nil
+  end
+
+  def sha256_certificate_hash
+    test_certificate ? test_certificate[:hashSha256] : nil
+  end
+
+  def empty_certs?
```
jladieu, line 37: Possible bug?  All three certs need to be missing for this to evaluate to true.  Shouldn't this be true if _any_ are nil?

And a nit: could this be renamed to `certs_provided?` which evaluates to true when all three are non-nil and non-empty?  This will require you to invert some logic where `empty_certs?` gets used, but will let you get rid of the double-negative `unless` [clause](https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/pull/219/files#diff-1f50a993e035d5ff62885ac3fddac664536c4ca69bb273ea666d890d3b2b5621R42)

tonybaroneee: Nope, the backend either returns all or nothing. But I like `certs_provided?`, will update.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -84,9 +95,9 @@ def post_notes(app_id, release_id, release_notes)
       #   app_id - Firebase App ID
       #
       # Throws a user_error if the app hasn't been onboarded to App Distribution
-      def get_app(app_id, app_view = 'BASIC')
+      def get_app(app_id)
         begin
-          response = connection.get("#{v1_apps_url(app_id)}?appView=#{app_view}") do |request|
+          response = connection.get("#{v1alpha_apps_url(app_id)}?appView=BASIC") do |request|
```
jladieu, line 100: I'm confused about the need to change from `v1` to `v1alpha` here.

tonybaroneee: `GetApp` is a v1alpha endpoint (not part of the new API).

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -25,49 +26,59 @@ def initialize(auth_token, debug = false)
...
         begin
-          connection.post(release_notes_create_url(app_id, release_id), payload.to_json) do |request|
+          payload = {
+            release: {
+              name: release_name(project_number, app_id, release_id),
+              releaseNotes: {
+                text: release_notes
+              }
+            },
+            updateMask: "release_notes.text"
```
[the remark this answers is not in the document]
tonybaroneee: It lets the backend know which field of the resource is being updated. More at https://google.aip.dev/161

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -198,33 +226,59 @@ def get_udids(app_id)
...
 
-      def upload_status_url(app_id, app_token)
-        "#{v1_apps_url(app_id)}/upload_status/#{app_token}"
+      def upload_status_url(operation_name)
+        "/v1/#{operation_name}"
       end
 
       def get_udids_url(app_id)
-        "#{v1_apps_url(app_id)}/testers:getTesterUdids"
+        "#{v1alpha_apps_url(app_id)}/testers:getTesterUdids"
```
jladieu, line 262: I didn't realize this was already supported in the old API -- to me this would dictate adding it to our build plan for GA.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -96,16 +107,37 @@ def get_app(app_id, app_view = 'BASIC')
...
+      # Get AAB info (Android apps only)
+      #
+      # args
+      #   project_number - Firebase project number
+      #   app_id - Firebase App ID
+      #
+      # Throws a user_error if the app hasn't been onboarded to App Distribution
+      def get_aab_info(project_number, app_id)
+        begin
+          aab_name = aab_info_name(project_number, app_id)
```
jladieu, line 119: This looks like it might be double duty...isn't `aab_info_url(project_number, app_id)` == `aab_name`? - why do you need to pass in a name param here if that's what the URL already is?

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
```diff
@@ -3,35 +3,50 @@ def initialize(response_json_hash)
...
 
-  def release_hash
-    @response_json_hash[:release]
+  def error
+    @response_json_hash[:error]
   end
 
-  def release_id
-    release_hash ? release_hash[:id] : nil
+  def message
```
jladieu, line 33: Is this ever populated when it's _not_ an error?

If this is strictly an error message, can it be renamed to `error_message`?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: True, might be nice to pull it before releasing a new version if we wanted to rid this code of v1alpha references entirely (assuming we also get rid of `GetApp` in favor of app id parsing for project number).

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
tonybaroneee: Agreed, but I didn't want to throw an internal google URL in this public code. We'll know to update this code if that enum is ever changed (we'd scrub through all of our client code in that case).

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
jladieu: This might be worth a higher bandwidth conversation because I'm confused now :)

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
jladieu: Ok but in the surprising event if only one of the three are nil, wouldn't you want to also crash & burn there?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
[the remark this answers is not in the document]
jladieu: Neat, thanks for sharing!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Nope, they are different (one is prefixed with `/v1/`.

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tonybaroneee: Updated to parse project number from `app_id` and removed `GetApp` entirely.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -78,22 +86,23 @@ def post_notes(app_id, release_id, release_notes)
...
       # args
       #   app_id - Firebase App ID
       #
       # Throws a user_error if the app hasn't been onboarded to App Distribution
-      def get_app(app_id, app_view = 'BASIC')
+      def get_aab_info(app_id)
         begin
-          response = connection.get("#{v1_apps_url(app_id)}?appView=#{app_view}") do |request|
+          aab_name = aab_info_name(app_id)
+          response = connection.get("#{aab_info_url(app_id)}?name=#{aab_name}") do |request|
```
tagboola, line 98: I don't _think_ you need the `name` query parameter. The backend should be able to extract it from the URL.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -105,7 +114,7 @@ def get_app(app_id, app_view = 'BASIC')
       #
       # Throws a user_error if the binary file does not exist
       def upload_binary(app_id, binary_path, platform)
-        connection.post(binary_upload_url(app_id), read_binary(binary_path)) do |request|
+        response = connection.post(binary_upload_url(app_id), read_binary(binary_path)) do |request|
```
tagboola, line 117: Should we be setting the internally visible `client_info` field here instead of the headers?

https://source.corp.google.com/piper///depot/google3/google/firebase/appdistro/v1/firebase_app_distribution.proto;l=216;rcl=389870755

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
```diff
@@ -0,0 +1,40 @@
...
+  def sha1_certificate_hash
+    test_certificate[:hashSha1]
+  end
+
+  def sha256_certificate_hash
+    test_certificate[:hashSha256]
+  end
+
+  def certs_provided?
+    !!md5_certificate_hash && !!sha1_certificate_hash && !!sha256_certificate_hash
```
jladieu, line 38: I believe this has a slightly different meaning than what the prior code was...this is _only_ a nilness check on all three, so it will treat empty hashes as `provided`

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -123,57 +134,50 @@ def upload_binary(app_id, binary_path, platform)
...
-              break
-            elsif upload_status_response.in_progress?
-              sleep(POLLING_INTERVAL_SECONDS)
+        UI.message("⌛ Uploading the #{binary_type}.")
+        operation_name = upload_binary(app_id, binary_path, platform)
+
+        upload_status_response = get_upload_status(operation_name)
+        MAX_POLLING_RETRIES.times do
+          if upload_status_response.success? || upload_status_response.already_uploaded?
+            UI.success("✅ Uploaded the #{binary_type}.")
```
tagboola, line 153: Ooooooh could we update this message to specify where the binary resulted in a new release or if it was grouped in an existing release?

jladieu (approved):

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
```diff
@@ -3,35 +3,50 @@ def initialize(response_json_hash)
...
   end
 
-  def release_id
-    release_hash ? release_hash[:id] : nil
+  def error_message
+    error ? error[:message] : nil
+  end
+
+  def success?
+    done && status == 'RELEASE_CREATED'
```
tagboola, line 38: Wouldn't it also be successful if `status = 'RELEASE_UPDATED'`?

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
tonybaroneee: Good call, updated to the previous double protection, just negated.

tagboola (approved):

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
tonybaroneee: Good catch, I updated this to just check for a non-nil `release` instead.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Great call! Added two more cases for `RELEASE_UNMODIFIED` and `RELEASE_UPDATED`.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: We decided to drop the `client_info` param and stick with headers. That was only introduced when we thought we couldn't set headers using the new ts_api_client call in fireconsole, but we're going back to making a vanilla POST call.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Good call, removed.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -29,32 +29,34 @@ def self.run(params)
...
-        app = fad_api_client.get_app(app_id, app_view)
-        validate_app!(app, binary_type)
-        release_id = fad_api_client.upload(app.project_number, app_id, binary_path, platform.to_s)
+        # If binary is an AAB, get the AAB info for this app, which includes the integration state and certificate data
+        if binary_type == :AAB
+          aab_info = fad_api_client.get_aab_info(app_id)
+          validate_aab_setup!(aab_info)
+        end
+
+        release_id = fad_api_client.upload(app_id, binary_path, platform.to_s)
```
kaibolay, line 38: would be nice if this would return the entire `release_name` which than can be passed to `update_release_notes` and `enable_access` (which I think should be renamed to `distribute`)

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -29,32 +29,34 @@ def self.run(params)
...
-        fad_api_client.post_notes(app_id, release_id, release_notes(params))
+        fad_api_client.update_release_notes(app_id, release_id, release_notes(params))
 
         testers = get_value_from_value_or_file(params[:testers], params[:testers_file])
         groups = get_value_from_value_or_file(params[:groups], params[:groups_file])
         emails = string_to_array(testers)
-        group_ids = string_to_array(groups)
-        fad_api_client.enable_access(app_id, release_id, emails, group_ids)
+        group_aliases = string_to_array(groups)
+        fad_api_client.enable_access(app_id, release_id, emails, group_aliases)
```
kaibolay, line 59: rename `enable_access` to `distribute` or `distribute_release`?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -29,32 +29,34 @@ def self.run(params)
...
         fad_api_client = Client::FirebaseAppDistributionApiClient.new(auth_token, params[:debug])
 
-        # If binary is an AAB get FULL view of app which includes the aab_state
-        app_view = binary_type == :AAB ? 'FULL' : 'BASIC'
-        app = fad_api_client.get_app(app_id, app_view)
-        validate_app!(app, binary_type)
-        release_id = fad_api_client.upload(app.project_number, app_id, binary_path, platform.to_s)
+        # If binary is an AAB, get the AAB info for this app, which includes the integration state and certificate data
+        if binary_type == :AAB
+          aab_info = fad_api_client.get_aab_info(app_id)
```
kaibolay, line 34: would be nice if all methods would be changed to work with resource names rather than IDs, so `app_id` would be replaced by `app_name` and `app_id`+`release_id` would just be `release_name`

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
kaibolay: Maybe change this message to explicitly point out that a new release was created?

Also: How about including the `display_version` and `build_version` in all three of these messages?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -123,57 +133,56 @@ def upload_binary(app_id, binary_path, platform)
...
-          end
-          MAX_POLLING_RETRIES.times do
-            upload_status_response = get_upload_status(app_id, upload_token)
-            if upload_status_response.success? || upload_status_response.already_uploaded?
-              UI.success("✅ Uploaded the #{binary_type}.")
-              break
-            elsif upload_status_response.in_progress?
-              sleep(POLLING_INTERVAL_SECONDS)
+        UI.message("⌛ Uploading the #{binary_type}.")
+        operation_name = upload_binary(app_id, binary_path, platform)
```
kaibolay, line 147: The old code calls `get_upload_status` before attempting a new upload as an optimization to detect duplicate uploads.

To restore this functionality, your new code needs to call `get_operation_name` and `get_upload_status`  before calling `upload_binary`.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -198,33 +207,54 @@ def get_udids(app_id)
 
       private
 
-      def v1_apps_url(app_id)
+      def project_number(app_id)
+        app_id.split(':')[1]
+      end
+
+      def v1alpha_apps_url(app_id)
```
kaibolay, line 214: I think all these methods should take resource names (e.g. `app_name` and `release_name`) as parameters instead of IDs.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -198,33 +207,54 @@ def get_udids(app_id)
...
       def get_udids_url(app_id)
-        "#{v1_apps_url(app_id)}/testers:getTesterUdids"
+        "#{v1alpha_apps_url(app_id)}/testers:getTesterUdids"
+      end
+
+      def app_name(app_id)
+        "projects/#{project_number(app_id)}/apps/#{app_id}"
+      end
+
+      def release_name(app_id, release_id)
```
kaibolay, line 250: this method should not be needed

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
```diff
@@ -3,35 +3,54 @@ def initialize(response_json_hash)
...
+
+  def response
+    @response_json_hash[:response]
+  end
+
+  def release
+    response ? response[:release] : nil
+  end
+
+  def release_id
```
kaibolay, line 18: this method should not be needed once working with resource names

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tonybaroneee: Good idea. I like `distribute` since it's parallel to `upload`.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Done

lib/fastlane/plugin/firebase_app_distribution/helper/upload_status_response.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Done.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: I think we'd have to handle a 404 from `get_upload_status`, so it would be a little more complicated. Although I do see the appeal in saving an upload if it's not needed.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tonybaroneee: Done!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Done!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -198,33 +205,45 @@ def get_udids(app_id)
...
+      def upload_status_url(operation_name)
+        "/v1/#{operation_name}"
       end
 
       def get_udids_url(app_id)
-        "#{v1_apps_url(app_id)}/testers:getTesterUdids"
+        "#{v1alpha_apps_url(app_id)}/testers:getTesterUdids"
+      end
+
+      def app_name(app_name)
```
kaibolay, line 240: ?

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -198,33 +205,45 @@ def get_udids(app_id)
...
+        "projects/#{project_number(app_name)}/apps/#{app_id}"
       end
 
-      def get_upload_token(project_number, app_id, binary_path)
+      def get_operation_name(app_name, binary_path)
         binary_hash = Digest::SHA256.hexdigest(read_binary(binary_path))
-        CGI.escape("projects/#{project_number}/apps/#{app_id}/releases/-/binaries/#{binary_hash}")
+        CGI.escape("projects/#{project_number(app_name)}/apps/#{app_id}/releases/-/operations/#{binary_hash}")
```
kaibolay, line 246: I don't think there should be any escaping - and it's just 
`"#{app_name}/releases/-/operations/#{binary_hash}"`

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Oops, removed.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Yeah agreed, not sure why it was there to behind with.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -123,57 +131,56 @@ def upload_binary(app_id, binary_path, platform)
...
+        UI.message("⌛ Uploading the #{binary_type}.")
+        operation_name = upload_binary(app_name, binary_path, platform)
+
+        upload_status_response = get_upload_status(operation_name)
+        MAX_POLLING_RETRIES.times do
+          if upload_status_response.success?
+            UI.success("✅ Uploaded the #{binary_type} and created release #{upload_status_response.release_version}.")
+            break
+          elsif upload_status_response.release_updated?
+            UI.success("✅ The same #{binary_type} was found, and the existing #{upload_status_response.release_version} release has been updated.")
```
kaibolay, line 153: "same"might be misleading, since the provisioning profile changed. I don't know how explicit we want to be, but we could say something along the lines "✅ Uploaded #{binary_type} successfully; updated provisioning profile of existing release {upload_status_response.release_version}"

I guess we should agree on verbiage for all three CLI tools and make it identical (or at least very similar)

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
kaibolay: I think we probably want to keep that functionality (even though it breaks some abstraction, because we construct the operation name from the sha-256). So you'll need to add the error handling.

Or (alternatively) we decide we don't want that optimization any more - in which case we should remove it from all three CLI clients.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: This won't be the case for android. This can happen if they add testers to an existing release, for example.

kaibolay: No, not really. On Android there's always exactly one binary per release. So either we get `RELEASE_CREATED` or `RELEASE_UNMODIFIED`. The `RELEASE_UPDATED` can only happen on iOS when the "instance identifier" (http://google3/java/com/google/firebase/appdistro/boq/appmetadataextractor/service/ios/IpaMetadata.java?l=63-73&rcl=390029034) remains the same, but the sha-256 of the entire binary changes.

For Android the instance identifier is the sha-1 of the entire binary. So if the binary changes, both the sha-1 and sha-256 change and a new release is created. If the binary doesn't change, it's a duplicate upload.

tonybaroneee: Sorry, I know about the iOS provisioning profile case. I just realized that what I was referring to is unrelated to upload and more of an enable access concern.

lib/fastlane/plugin/firebase_app_distribution/client/aab_info.rb
jladieu: I agree about not throwing an internal google URL in this public code.  To clarify my suggestion, it was to explicitly grab a link to _this_ code and drop it into the source file in codesearch.  I trust that you would scrub through the client code, but if I were making a change to the enum, I wouldn't know where to look.

jladieu: You won't be able to actually do this until after this is merged though since aab_info.rb isn't on the main branch yet :)

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
kaibolay: I've chatted with @lfkellogg and we agreed to remove this optimization. So you can remove `get_operation_name` since it won't be needed. I'll also remove this functionality from the Firebase CLI and we should remove it from gradle as well...

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Updated to use your wording instead for this (I dig it!)

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tonybaroneee: Sounds good, done.

kaibolay (approved):

PR #222, opened 2021-08-16
Add AddTesters and RemoveTesters actions

tagboola:


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -20,6 +21,8 @@ module ErrorMessage
   APP_NOT_PUBLISHED = "This app is not published in the Google Play console."
   NO_APP_WITH_GIVEN_BUNDLE_ID_IN_PLAY_ACCOUNT = "App with matching package name does not exist in Google Play."
   PLAY_IAS_TERMS_NOT_ACCEPTED = "You must accept the Play Internal App Sharing (IAS) terms to upload AABs."
+  INVALID_EMAIL_ADDRESS = "You passed in invalid email address."
```
lfkellogg, line 24: "an" invalid email address?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
```diff
@@ -0,0 +1,78 @@
...
+          FastlaneCore::ConfigItem.new(key: :project,
+                                      env_name: "FIREBASEAPPDISTRO_PROJECT_NUMBER",
+                                      description: "Your Firebase project number. You can find the project number in the Firebase console, on the General Settings page",
+                                      type: Integer),
+          FastlaneCore::ConfigItem.new(key: :emails,
+                                      env_name: "FIREBASEAPPDISTRO_ADD_TESTERS_EMAILS",
+                                      description: "Comma separated list of tester emails to be created. A maximum of 1000 testers can be created at a time",
+                                      optional: true,
+                                      type: String),
+          FastlaneCore::ConfigItem.new(key: :file,
```
lfkellogg, line 56: How about `:emails_file` to match the convention from the main action?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        emails = get_value_from_value_or_file(params[:emails], params[:file]).split(',')
+
+        if emails.count > 1000
+          UI.user_error!("A maximum of 1000 testers can be added at a time.")
+        end
+
+        fad_api_client.add_testers(params[:project], emails)
+
+        UI.success("✅ Tester(s) successfully added.")
```
lfkellogg, line 29: Even though it's not as meaningful as in the delete case, but how about getting the count in there just for reference, e.g. `"#{emails.count} tester(s) successfully added."`?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
```diff
@@ -0,0 +1,78 @@
...
+        "Create testers"
+      end
+
+      def self.authors
+        ["Tunde Agboola"]
+      end
+
+      # supports markdown.
+      def self.details
+        "Create testers"
```
lfkellogg, line 42: Maybe some more details here, like "Create testers in bulk from a comma-separated list or a file." or even more detail? Maybe a link to docs if there is one?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
```diff
@@ -0,0 +1,79 @@
...
+        "Delete testers"
+      end
+
+      def self.authors
+        ["Tunde Agboola"]
+      end
+
+      # supports markdown.
+      def self.details
+        "delete testers"
```
lfkellogg, line 42: Same here - more details? (Also capitalization on this one)

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -44,6 +44,14 @@ def get_ios_app_id_from_archive_plist(archive_path, plist_path)
         UI.shell_error!("can't extract GOOGLE_APP_ID") if identifier.empty?
         return identifier
       end
+
+      def blank?(value)
+        value.respond_to?(:empty) ? !!value.empty? : !value
```
lfkellogg, line 49: I think this needs to be `value.respond_to?(:empty?)` (note the question mark in the symbol). I tested this out in irb:

```
 ~  irb
irb(main):001:0> [].respond_to?(:empty?)
=> true
irb(main):002:0> [].respond_to?(:empty)
=> false
```

lfkellogg: Also you shouldn't have to use `!!` to coerce the result of `.empty?` to a boolean since AFAICT it only returns true or false: https://apidock.com/ruby/Array/empty%3F

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
```diff
@@ -44,6 +44,14 @@ def get_ios_app_id_from_archive_plist(archive_path, plist_path)
...
+      end
+
+      def present?(value)
```
lfkellogg, line 52: This is unused. Did you want to include it for future use anyway? It does seem helpful, I know we've got some ugly presence-checking scattered around this codebase.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -441,4 +441,95 @@
         .to raise_error("#{ErrorMessage::INVALID_TESTERS} \nEmails: #{emails} \nGroups: #{group_ids}")
     end
   end
+
+  describe '#add_testers' do
+    let(:headers) { { 'Authorization' => 'Bearer auth_token', 'Content-Type' => 'application/json' } }
+
+    it 'is successful' do
+      emails = %w[1@foo.com 2@foo.com]
```
lfkellogg, line 449: TIL `%w` is a thing 🤯

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -441,4 +441,95 @@
...
+      end
+      expect { api_client.add_testers("project_number", emails) }
+        .to raise_error(ErrorMessage::TESTER_LIMIT_VIOLATION)
+    end
+  end
+
+  describe '#remove_testers' do
+    let(:headers) { { 'Authorization' => 'Bearer auth_token', 'Content-Type' => 'application/json' } }
+
+    it 'is successful' do
```
lfkellogg, line 507: How about `'returns the number of testers successfully removed'`

spec/firebase_app_distribution_remove_testers_action_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it 'succeeds and makes call with value from emails param' do
+      project = 1
+      emails = '1@e.mail,2@e.mail'
+      path = 'path/to/file'
+      count = 1
+      fake_file = double('file')
+      allow(File).to receive(:open)
+        .with(path)
+        .and_return(fake_file)
+      allow(fake_file).to receive(:read).and_return('')
```
lfkellogg, line 32: Instead of allowing this read, can we expect the file not to receive `:read` in this case? Or just not mock out any of this file related stuff.

spec/firebase_app_distribution_add_testers_action_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    it 'succeeds and makes call with value from emails param' do
+      project = 1
+      emails = '1@e.mail,2@e.mail'
+      path = 'path/to/file'
+      fake_file = double('file')
+      allow(File).to receive(:open)
+        .with(path)
+        .and_return(fake_file)
+      allow(fake_file).to receive(:read).and_return('')
```
lfkellogg, line 30: Instead of allowing this read, can we expect the file not to receive `:read` in this case? Or just not mock out any of this file related stuff.

lfkellogg (commented):
A few pretty minor suggestions/comments. This looks awesome!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
tagboola: We originally suggested something like that i.e. `testers_file` that but the API council thought the less verbose `file` would be better

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
tagboola: Thought about this, but we technically wouldn't be able to know exactly how many testers were actually created. The response contains all the testers from the request even if they exist already, so there could be testers in there that weren't actually created.

lfkellogg: Got it. Could we add that note as a comment to make sure this message stays vague?

spec/firebase_app_distribution_add_testers_action_spec.rb
tagboola: Good call! Done.

spec/firebase_app_distribution_remove_testers_action_spec.rb
tagboola: Good call! Done.

spec/firebase_app_distribution_api_client_spec.rb
tagboola: Done.

spec/firebase_app_distribution_api_client_spec.rb
tagboola: Ha yeah rubocop was adamant that I use it 😆

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
tagboola: Yeah just kept it in for future use

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
tagboola: Good catch. I actually got this code from the rails source https://apidock.com/rails/Object/blank%3F. Added a comment specifying where it's from and also removed the coercion.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
tagboola: Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
tagboola: Done.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
tagboola: Done

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
tagboola: We originally suggested something like this i.e. `testers_file` but the API council thought the less verbose `file` would be better.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
tagboola: Considered this but in the response we include testers regardless of whether or not they were created or already existed so it technically wouldn't be an accurate count of then number of testers that were actually created.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -196,6 +196,51 @@ def get_udids(app_id)
...
+        connection.post(add_testers_url(project_number), payload.to_json) do |request|
+          request.headers[AUTHORIZATION] = "Bearer " + @auth_token
+          request.headers[CONTENT_TYPE] = APPLICATION_JSON
+        end
+      rescue Faraday::BadRequestError
+        UI.user_error!(ErrorMessage::INVALID_EMAIL_ADDRESS)
+      rescue Faraday::ResourceNotFound
+        UI.user_error!(ErrorMessage::INVALID_PROJECT)
+      rescue Faraday::ClientError => e
+        if e.response_status == 429
```
rebehe, line 217: Does the FE actually do tester limit checking?

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -441,4 +441,95 @@
...
+  describe '#add_testers' do
+    let(:headers) { { 'Authorization' => 'Bearer auth_token', 'Content-Type' => 'application/json' } }
+
+    it 'is successful' do
+      emails = %w[1@foo.com 2@foo.com]
+      stubs.post("/v1/projects/project_number/testers:batchAdd", { emails: emails }.to_json, headers) do |env|
+        [
+          200,
+          {},
+          {}
```
rebehe, line 454: Can you add comments explaining what these `{}` are?

rebehe (approved):

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
rebehe: Sorry, I meant BE*

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
lfkellogg: @rebehe That wouldn't be accurate either, since the project could already have other testers that were not included in the request.

@tagboola Your comment LGTM, thanks.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
lfkellogg: @tagboola you updated `self.description` but not `self.details` :)

I think if one is going to be more detailed it should be `details`.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
lfkellogg: Same thing here as above, should update `self.details`

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_helper.rb
lfkellogg: SGTM

spec/firebase_app_distribution_api_client_spec.rb
lfkellogg: Oh good to know!

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: Yep we check for tester limits on the backend

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
tagboola: Ahh my bad. Updated `details` as well.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
tagboola: Ahh my bad. Updated `details` as well.

spec/firebase_app_distribution_api_client_spec.rb
tagboola: Done!

lfkellogg (approved):

PR #236, opened 2021-08-20
Make project number required and add messages to actions

tagboola:


spec/firebase_app_distribution_remove_testers_action_spec.rb
```diff
@@ -9,14 +9,19 @@
     end
 
     it 'raises an error if emails and file are blank' do
-      expect { action.run({}) }
+      expect { action.run({ project_number: 1 }) }
+        .to raise_error("Must specify `emails` or `file`.")
+    end
+
+    it 'raises an error if emails and file are blank' do
```
lfkellogg, line 16: This is the same as the previous test case. I assume you meant to add a test where the `project_number` is not set?

lfkellogg (approved):

spec/firebase_app_distribution_remove_testers_action_spec.rb
tagboola: Ahh yeah my bad. I actually tried to run a test where project number isn't set but the actual parameter validation seems to be executed before `run` is called.  Deleted the duplicate test!

lfkellogg: Got it, thanks!

PR #238, opened 2021-08-24
Handle empty response in remove_testers call

tagboola:


spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -545,6 +545,21 @@
       expect(result).to eq(1)
     end
 
+    it 'returns the 0 if response is empty' do
```
rebehe, line 548: super nit: "returns the 0" -> "returns 0"

rebehe (approved):

tonybaroneee (approved):

lfkellogg (approved):

spec/firebase_app_distribution_api_client_spec.rb
tagboola: Good catch! Done.

PR #240, opened 2021-08-24
Handle whitespaces and empty email arrays

tagboola:
Prefer `string_to_array` instead of `split` since it handles whitespaces and handle case where `nil` or `[]` is returned from `string_to_array`


lfkellogg (approved):
Nice!

Issue #178, opened 2020-11-17
App Distribution could not generate credentials from the refresh token specified.


Robotex:
### Issue Description

Tried to setup firebase on a new development machine, I followed the login procedure using the plugin action firebase_app_distribution_login, pasted the login token and launched the build. On firebase_app_distribution step it throws 
> App Distribution could not generate credentials from the refresh token specified.

**Using the login token generated from firebase-cli it works correctly.**
* I noticed the plugin asks different permissions than firebase-cli (the latter asks additional permissions to show and manage firebase projects, whereas the plugin asks only to access google cloud platform services), maybe this is the problem?

##### Configuration

googleservice_info_plist_path

##### Environment

| Library     | Version |
| ----------- | ----------- |
| fastlane | 2.167.0 |
| fastlane-plugin-firebase_app_distribution | 0.2.4 |

lfkellogg:
Hi @Robotex - thanks for the report.

Just to make sure, can you confirm that you set your `FIREBASE_TOKEN` environment variable to the exact string that was printed out in your terminal after pasting in the code from your browser?

This value from when you ran `firebase_app_distribution_login`:
```
$ bundle exec firebase run firebase_app_distribution_login
...
[13:05:15]: Set the refresh token as the FIREBASE_TOKEN environment variable
[13:05:15]: Refresh Token: <THIS VALUE>
[13:05:15]: Result: true
...
```

Should be equal to this value when running `echo $FIREBASE_TOKEN`:
```
$ echo $FIREBASE_TOKEN
<THIS VALUE>
```

Robotex:
Hi Ifkellog,
You are right, I checked again and the value was incorrectly copied and pasted. My mistake!

Issue closed.

uguraktas:
> Hi @Robotex - thanks for the report.
> 
> Just to make sure, can you confirm that you set your `FIREBASE_TOKEN` environment variable to the exact string that was printed out in your terminal after pasting in the code from your browser?
> 
> This value from when you ran `firebase_app_distribution_login`:
> 
> ```
> $ bundle exec firebase run firebase_app_distribution_login
> ...
> [13:05:15]: Set the refresh token as the FIREBASE_TOKEN environment variable
> [13:05:15]: Refresh Token: <THIS VALUE>
> [13:05:15]: Result: true
> ...
> ```
> 
> Should be equal to this value when running `echo $FIREBASE_TOKEN`:
> 
> ```
[...]

PR #263, opened 2022-03-21
Improve error messages when failing to get access token

lfkellogg:
Hoping to help folks resolve their own refresh token issues (see https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/issues/178). This change:
- Prompts developers to set the debug parameter to true to get more info
- In debug mode, prints out the refresh token used

Example failure using service credential file, with `debug: false`:
```
[17:40:57]: App Distribution could not generate credentials from the service credentials file specified: "/path/to/bad.json". For more information, try again with firebase_app_distribution's "debug" parameter set to "true".
```

... and `debug: true`:
```
[18:08:09]: App Distribution could not generate credentials from the service credentials file specified: "/path/to/bad.json"
Authorization failed.  Server message:
{"error":"invalid_grant","error_description":"Invalid JWT Signature."}
Response status: 400
```

Example failure using refresh token, with `debug: false`:
```
[17:24:06]: App Distribution could not generate credentials from the refresh token specified. For more information, set firebase_app_distribution's "debug" parameter to "true".
```

...and `debug: true`:
```
[17:23:25]: App Distribution could not generate credentials from the refresh token specified.
Refresh token used: "badtoken"
Authorization failed.  Server message:
{
  "error": "invalid_grant",
  "error_description": "Bad Request"
}
Response status: 400
```

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -18,6 +18,10 @@ module Fastlane
       def self.run(params)
         params.values # to validate all inputs before looking for the ipa/apk/aab
 
+        if params[:debug]
+          UI.important("Warning: Debug logging enabled. Output may include sensitive information.")
+        end
+
         app_id = app_id_from_params(params)
         app_name = app_name_from_app_id(app_id)
         platform = lane_platform || platform_from_app_id(app_id)
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
@@ -113,12 +113,13 @@ module Fastlane
 
       # Formats and redacts a token for printing out during debug logging. Examples:
       #   'abcd' -> '"abcd"''
-      #   'abcdef1234' -> '"abcdeXXXXX" (redacted)'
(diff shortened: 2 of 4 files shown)


kaibolay (approved):
I guess it'd be a security risk to always log the refresh token, so that's why you only log it with the `debug` parameter

tagboola (commented):
Nice! Any worries about printing out the refresh token? Thought: Maybe we should printing out a warning when debug is set to true to be explicit that the output may contain sensitive information.

lfkellogg:
@tagboola A warning is probably a good idea regardless. In this case I guess the refresh token is particularly sensitive. What would you think about jut printing out the first few (5?) characters of it? I'm trying to help in the case where the token is empty or obviously invalid.

tagboola (commented):
@lfkellogg I like the idea of only printing the first few characters of the refresh token 🙂

kaibolay:
I think with `debug` it's appropriate to log raw data without trying to obfuscate potentially sensitive data. Don't we log the raw HTTP requests and responses? In that case the refresh token is exposed anyway... I think a warning is enough.

tagboola (commented):
I actually don't think we currently log the refresh token when `debug` is set to `true`. We log all of the requests we make to our API, but the we don't log the headers. Even if we did log the headers, that would only contain the short-lived access token, which isn't as bad as logging the long-lived refresh token.

lfkellogg:
Done, thanks for the good discussion folks 👍

jladieu (approved):
Great work!

Issue #266, opened 2022-04-06
Whitespaces are deleted from environment variable


alejandro-oria:
### Issue Description
I have a variable declared in my .env files which is the following:
```
FIREBASE_DISTRIBUTION_GROUPS = "iOS Team"
```

Once I run one of my lanes I get an error when trying to distribute the release to the testers since the distribution group doesn't exist. The thing is that the whitespaces are deleted from the variable's value. 
The desired value for the groups would be "iOS Team" (like it's declared in the .env files) instead of "iOSTeam".

##### Configuration

```
firebase_app_distribution(
      app: "[...]",
      googleservice_info_plist_path: "[...]/GoogleService-Info.plist",
      groups: ENV['FIREBASE_DISTRIBUTION_GROUPS'],
      release_notes: options[:release_notes],
      ipa_path: "./builds/[...].ipa",
      debug: true
    )
```
    

##### Complete verbose output
```
[13:44:01]: ---------------------------------------
[13:44:01]: --- Step: firebase_app_distribution ---
[13:44:01]: ---------------------------------------
[13:44:01]: Warning: Debug logging enabled. Output may include sensitive information.
[13:44:01]: Authenticating with FIREBASE_TOKEN environment variable
[13:44:01]: 🔐 Authenticated successfully.
[13:44:01]: ⌛ Uploading the IPA.
D, [2022-04-06T13:44:01.905910 #19909] DEBUG -- request: POST https://firebaseappdistribution.googleapis.com/upload/v1/projects/[...]
D, [2022-04-06T13:44:22.235495 #19909] DEBUG -- response: Status 200
D, [2022-04-06T13:44:22.235588 #19909] DEBUG -- response: {
  "name": "[...]"
}

D, [2022-04-06T13:44:22.235914 #19909] DEBUG -- request: GET https://firebaseappdistribution.googleapis.com/v1/projects/[...]
D, [2022-04-06T13:44:22.614706 #19909] DEBUG -- response: Status 200
D, [2022-04-06T13:44:22.614777 #19909] DEBUG -- response: {
  "name": "[...]",
  "done": true,
  "response": {
    "@type": "type.googleapis.com/google.firebase.appdistro.v1.UploadReleaseResponse",
    "result": "RELEASE_UNMODIFIED",
    "release": {
      "name": "[...]",
      "releaseNotes": {
        "text": "[...]"
      },
      "displayVersion": "[...]",
      "buildVersion": "[...]",
[...]

PR #267, opened 2022-04-11
Trim individual values from comma separate strings

tagboola:
In response to #266 we should only _trim_ the whitespaces from the individual values in a comma separated string

### DIFF
--- a/spec/firebase_app_distribution_helper_spec.rb
+++ b/spec/firebase_app_distribution_helper_spec.rb
@@ -61,8 +61,13 @@ describe Fastlane::Helper::FirebaseAppDistributionHelper do
     end
 
     it 'returns an array when the string passed in has multiple values seperated by commas' do
-      array = helper.string_to_array("string1, string2, string3")
-      expect(array).to eq(["string1", "string2", "string3"])
+      array = helper.string_to_array("string1,string2,string3")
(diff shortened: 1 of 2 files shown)


lfkellogg (approved):
Nice!

spec/firebase_app_distribution_helper_spec.rb
```diff
@@ -61,8 +61,13 @@
...
 
     it 'returns an array when the string passed in has multiple values seperated by commas' do
-      array = helper.string_to_array("string1, string2, string3")
-      expect(array).to eq(["string1", "string2", "string3"])
+      array = helper.string_to_array("string1,string2,string3")
+      expect(array).to eq(%w[string1 string2 string3])
+    end
+
+    it 'returns an array with trimmed values' do
+      array = helper.string_to_array("string1, str ing2 , string3")
```
kaibolay, line 69: also add leading and trailing spaces? And multiple spaces? E.g. something like `"   string1,   str ing2    , string3  "`

kaibolay:
Better error message in separate PR?

spec/firebase_app_distribution_helper_spec.rb
tagboola: Great suggestion. Updated!

tagboola:
Yep will update the error message in a separate PR

PR #268, opened 2022-04-11
Be more explicit about group aliases

tagboola:
Update parameter descriptions and improve invalid testers error message


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -11,7 +11,7 @@ module ErrorMessage
   INVALID_APP_ID = "App Distribution could not find your app. Make sure to onboard your app by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution. App ID"
   INVALID_PROJECT = "App Distribution could not find your Firebase project. Make sure to onboard an app in your project by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution."
   INVALID_PATH = "Could not read content from"
-  INVALID_TESTERS = "Could not enable access for testers. Check that the groups exist and the tester emails are formatted correctly"
+  INVALID_TESTERS = "Could not enable access for testers. Check that the tester emails are formatted correctly, the groups exist and the group alias is being used."
```
kaibolay, line 14: "[...] and the group alias is being used" --> "[...] and you are using group aliases (NOT group names) if specifying groups"

Even better: Check with Victoria

tagboola: Updated!

Issue #269, opened 2022-04-12
Can't login using firebase_app_distribution_login


RalfK92:
### Issue Description
When i try to use the firebase_app_distribuition_login function the link that is generated returns a 400 response code in the browser. I have tested it on my Macbook on our Mac Mini Build Server and a colleague at my company is having the issue.
<img width="535" alt="Screenshot 2022-04-12 at 11 16 19" src="https://user-images.githubusercontent.com/4293838/162926019-3df13ce5-f5dc-480a-843d-31faac83cff1.png">

##### Configuration

bundle exec fastlane run firebase_app_distribution_login --verbose

##### Complete verbose output
```
[✔] 🚀
DEBUG [2022-04-12 11:13:37.70]: Checking if there are any plugins that should be loaded...
DEBUG [2022-04-12 11:13:37.70]: Loading 'fastlane-plugin-appicon' plugin
INFO [2022-04-12 11:13:37.70]: gem 'fastlane-plugin-appicon' is already installed
DEBUG [2022-04-12 11:13:37.71]: Loading 'fastlane-plugin-firebase_app_distribution' plugin
INFO [2022-04-12 11:13:37.71]: gem 'fastlane-plugin-firebase_app_distribution' is already installed
+-------------------------------------------+---------+------------------------------------------------+
|                                             Used plugins                                             |
+-------------------------------------------+---------+------------------------------------------------+
| Plugin                                    | Version | Action                                         |
+-------------------------------------------+---------+------------------------------------------------+
| fastlane-plugin-appicon                   | 0.16.0  | android_appicon, appicon                       |
| fastlane-plugin-firebase_app_distribution | 0.3.3   | firebase_app_distribution_login,               |
|                                           |         | firebase_app_distribution_add_testers,         |
|                                           |         | firebase_app_distribution_get_udids,           |
|                                           |         | firebase_app_distribution,                     |
|                                           |         | firebase_app_distribution_get_latest_release,  |
|                                           |         | firebase_app_distribution_remove_testers       |
+-------------------------------------------+---------+------------------------------------------------+

objc[27891]: Class AppleTypeCRetimerRestoreInfoHelper is implemented in both /usr/lib/libauthinstall.dylib (0x1e52d5eb0) and /Library/Apple/System/Library/PrivateFrameworks/MobileDevice.framework/Versions/A/MobileDevice (0x1040904f8). One of the two will be used. Which one is undefined.
[...]

PR #270, opened 2022-04-12
Add loopback IP address to google login flow

tagboola:
The copy/paste method is being [deprecated](https://developers.google.com/identity/protocols/oauth2/native-app#manual-copypaste) and has been disabled for our client (see #269).

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
@@ -1,11 +1,10 @@
 require 'googleauth'
-require 'googleauth/stores/file_token_store'
+require "fileutils"
 require "fileutils"
 
 module Fastlane
   module Actions
     class FirebaseAppDistributionLoginAction < Action
-      OOB_URI = "urn:ietf:wg:oauth:2.0:oob"
       SCOPE = "https://www.googleapis.com/auth/cloud-platform"
 
       # In this type of application, the client secret is not treated as a secret.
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -55,6 +90,18 @@ def self.authors
       def self.is_supported?(platform)
         [:ios, :android].include?(platform)
       end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :port,
+                                       env_name: "FIREBASEAPPDISTRO_LOGIN_PORT",
+                                       description: "Port for the local web server which receives the response from Google's authorization server",
+                                       default_value: "8080",
```
tagboola, line 99: Do you think it's worth adding it logic to find a random open port?

lfkellogg: I think it would be fine to catch the error if the port is in use and recommend using the --port flag.

lfkellogg (approved):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -14,24 +13,60 @@ class FirebaseAppDistributionLoginAction < Action
...
-        UI.error("The code you entered is invalid. Copy and paste the code and try again.")
+        UI.success("Refresh Token: #{user_credentials.refresh_token}")
       rescue => error
         UI.error(error.to_s)
-        UI.crash!("An error has occured, please login again.")
+        UI.crash!("An error has occurred, please login again.")
+      end
+
+      def self.get_authorization_code(port)
+        server = TCPServer.open(port)
```
lfkellogg, line 47:
```suggestion
        begin
          server = TCPServer.open(port)
        rescue Errno::EADDRINUSE => error
          UI.error(error.to_s)
          UI.crash!("Port #{port} is in use. Please specify a different one using the --port option.")
        end
```

Issue #274, opened 2022-05-25
Update to accept access token param


enchorb:
Can the following line in `firebase_app_distribution_action.rb` be updated in the next version of this plugin.
I have this working locally and in our CI/CD, as we obtain an access token through WLIF (workload identity federation) and service account impersonation and pass in the access token as a param directly 
 
```
auth_token = params[:access_token] || fetch_auth_token(
          params[:service_credentials_file], params[:firebase_cli_token], params[:debug]
        )
```

mannyjimenez0810:
Hello @enchorb Thanks for reaching out about this. Could you tell me more about your use case? Is there some reason that the ```params[:service_credentials_file]``` doesn't work for you?

enchorb:
Hey, that works just fine but the reason for this is so we don't need to store any long-lived credentials like service accounts.

In CI/CD we use WLIF (workload identity federation) which is then able to impersonate a service account to get an access token which is then added to the `.env` that Fastlane and by extension this plugin would read the access token from

mannyjimenez0810:
Hey, are you able to use the ```GOOGLE_APPLICATION_CREDENTIALS``` environmental variable to authenticate? We already [support](https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/b2dd465d20f739edef9f66093298c425525f2c9a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb#L35) Application Default Credentials which is used in WIF. Does that work for your use case?

enchorb:
Hey @lfkellogg `GOOGLE_APPLICATION_CREDENTIALS` does not carry over jobs/workflows in CI/CD, a more common use case is to pass an impersonated access token through the app as a variable as so: 

```
     - name: Authenticate To Google Cloud
        if: github.event_name == 'push'
        uses: google-github-actions/auth@v0
        with:
          token_format: access_token
          access_token_lifetime: 300s
          access_token_scopes: https://www.googleapis.com/auth/cloud-platform
          workload_identity_provider: projects/${{ steps.environment.outputs.gcp_num }}/locations/global/workloadIdentityPools/gh-${{ github.event.repository.name }}-pool/providers/gh-${{ github.event.repository.name }}-provider
          service_account: gh-${{ github.event.repository.name }}-runner@${{ steps.environment.outputs.gcp_id }}.iam.gserviceaccount.com

      - name: Impersonate Deployment Account
[...]

PR #275, opened 2022-05-26
Adding in support for passing in an access token

mannyjimenez0810:


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -31,7 +31,7 @@ def self.run(params)
         UI.user_error!("Couldn't find binary at path #{binary_path}") unless File.exist?(binary_path)
         binary_type = binary_type_from_path(binary_path)
 
-        auth_token = fetch_auth_token(
+        auth_token = params[:access_token] || fetch_auth_token(
```
lfkellogg, line 34: The `access_token` param does not exist yet. You'd have to add it to `self.available_options` below.

mannyjimenez0810: Thanks!

tagboola (changes requested):
Hate to be the wet blanket here, but I think we this would require an API review. What worries me is that this is not a parameter that the Firebase CLI supports, even though it _seemingly_ supports WIF. Ideally, we want a consistent auth experience across _all_ CLI tools.

https://github.com/firebase/firebase-tools/issues/3926
https://tables.area120.google.com/table/8Ls4BVBK9Fn98JVbhvK_jQ/row/9CaYXMOtErA4El6aLw__6A

mannyjimenez0810:
The API review is unfortunate. Did we go through an initial review for the service_credentials_file? I tried looking for a api doc, but couldn't find one for that I found one for other parts.

Also why should we wait for Firebase CLI? If google support this auth shouldn't we also support this auth.  Looks like they updated their ruby auth to support this not too long ago https://github.com/googleapis/google-auth-library-ruby/issues/346. I did see some users also want Fastlane to have this feature, although that work was started months ago, it has not been picked up recently.

tagboola:
If the real problem is we need to support authentication with Workload Identity Federation, which is referenced in #274, I think we should try to do it in a way that's consistent with other clients.  Based on https://google.aip.dev/auth/4117, the guidance for authenticating with Workload Identity Federation is to generate a JSON configuration file that can then be used with [Application Default Credentials](https://google.aip.dev/auth/4110). The plugin already [supports](https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/b2dd465d20f739edef9f66093298c425525f2c9a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb#L35) Application Default Credentials, so why not try and support this way?

mannyjimenez0810:
@tagboola and I chatted offline. Unless this is customer blocking we will not pursue this any longer unless other CLI tools also update to support this authentication method. Happy to chat @lfkellogg if you had some thoughts or wanted some more context on this.

PR #277, opened 2022-07-25
Deprecate login action

tagboola:
Screenshot of proposed message when running `bundle exec fastlane run firebase_app_distribution_login`
Screen Shot 2022-07-27 at 3 56 51 PM

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
@@ -112,8 +112,8 @@ module Fastlane
       end
 
       def self.deprecated_notes
-        "The firebase_app_distribution_login task is deprecated and will be removed in a future release. See "\
-        "https://firebase.google.com/docs/app-distribution/android/distribute-gradle#step_2_authenticate_with_firebase "\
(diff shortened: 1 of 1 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
```diff
@@ -106,6 +106,16 @@ def self.available_options
 
         ]
       end
+
+      def self.category
+        :deprecated
+      end
```
lfkellogg, line 112: Just curious: what does this look like when you run `fastlane action firebase_app_distribution_login`? I assume there's some messaging there as well.

lfkellogg (approved):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_login.rb
tagboola: Looks like it prints the same message 🙂
Screen Shot 2022-07-25 at 1 52 17 PM

lfkellogg: Cool! Thanks.

VictoriaAuto (approved):

PR #282, opened 2022-09-27
Adding parsing error handling and improving error messages 

mannyjimenez0810:
Improving the error messaging when the tokens field is nil or missing.


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -48,20 +48,27 @@ def fetch_auth_token(google_service_path, firebase_cli_token, debug = false)
...
-
+        config_path = format_config_path
         if File.exist?(config_path)
-          begin
-            refresh_token = JSON.parse(File.read(config_path))['tokens']['refresh_token']
-            refresh_token unless refresh_token.nil? || refresh_token.empty?
-          # TODO: Catch parser errors, improve error handling here
-          # Returns nil when there is an empty "tokens" field in the firebase-tools json
-          rescue NoMethodError
+         begin
```
kaibolay, line 53: should the indent be two spaces instead of one?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -48,20 +48,27 @@ def fetch_auth_token(google_service_path, firebase_cli_token, debug = false)
...
+          firebase_tools_tokens = JSON.parse(File.read(config_path))['tokens']
+          if (firebase_tools_tokens.nil?)
+           UI.important("Unable to find \'tokens\' field in file located at #{config_path} Ensure that the file has a tokens field and try again")
+           return
           end
+          refresh_token = firebase_tools_tokens['refresh_token']
```
kaibolay, line 59: what if the 'refresh_token' is missing? I guess that should also produce a helpful error message. Or is that reported on a higher level if this function returns `nil`?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -48,20 +48,27 @@ def fetch_auth_token(google_service_path, firebase_cli_token, debug = false)
...
+         rescue JSON::ParserError
+          UI.important("Malformed json file located at #{config_path}")
+         end
+         refresh_token unless refresh_token.nil? || refresh_token.empty?
+        end
+      end
+
+      def format_config_path
```
kaibolay, line 67: nit: `get_config_path`?

kaibolay (commented):
can this be tested?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
mannyjimenez0810: good catch!

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
mannyjimenez0810: My first thought was get but i feel like I have gotten feedback that get is usually an API call or something of that nature. will update it since i like get!

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
mannyjimenez0810: If its nil thats fine we have a check on 63 that checks if its nil or empty. We just cant have tokens be nil or empty since we try to get the refresh token from it.

mannyjimenez0810:
> 
We do have tests to make sure if there is no tokens field and a test if the refresh token is also missing. 
https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/d99970830a2930ca0dd26ee5f65f7e71ede4cc74/spec/firebase_app_distribution_auth_client_spec.rb#L136

kaibolay:
> We do have tests to make sure if there is no tokens field and a test if the refresh token is also missing.
> 
> https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/d99970830a2930ca0dd26ee5f65f7e71ede4cc74/spec/firebase_app_distribution_auth_client_spec.rb#L136

Could those test be expanded to verify the new and improved error messges?

kaibolay (approved):

mannyjimenez0810:
> > We do have tests to make sure if there is no tokens field and a test if the refresh token is also missing.
> > https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/d99970830a2930ca0dd26ee5f65f7e71ede4cc74/spec/firebase_app_distribution_auth_client_spec.rb#L136
> 
> Could those test be expanded to verify the new and improved error messges?

We probably could but we don't typically test the extra information just the actual error messages. So in these cases I decided not to change the actual error message, but to add more information to those errors. It makes sense to make these actual error messages. I will make these new messages show as errors instead of extra text for the user and also check those on the tests. 

Screen Shot 2022-09-27 at 4 15 37 PM

kaibolay (approved):

PR #290, opened 2023-01-18
Add client version request header to all requests

rebehe:
Also tweak the text for the tester release link to be more clear.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -76,7 +76,7 @@ module Fastlane
         end
 
         if upload_status_response.testing_uri
-          UI.message("🔗 Share this release with testers: #{upload_status_response.testing_uri}")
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -76,7 +76,7 @@ def self.run(params)
         end
 
         if upload_status_response.testing_uri
-          UI.message("🔗 Share this release with testers: #{upload_status_response.testing_uri}")
+          UI.message("🔗 Share this release with testers that have access: #{upload_status_response.testing_uri}")
```
mannyjimenez0810, line 79: I believe you left a nit in the comments where "that"  was replaced with "who"

mannyjimenez0810 (approved):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
rebehe: Good call - done

PR #296, opened 2023-01-25
Create GitHub Action for publishing to RubyGems

rebehe:


tagboola:
Nice! Does this work without an OTP token?

tagboola (approved):

rebehe:
@tagboola You can enable MFA on a specific API key, so I think we're okay

PR #299, opened 2023-02-13
Save release to shared lane variable

tagboola:


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -62,14 +63,20 @@ def self.run(params)
           end
         end
 
-        fad_api_client.update_release_notes(release_name, release_notes(params))
+        release_notes = release_notes(params)
+        if release_notes.nil? || release_notes.empty?
+          UI.success("✅ No release notes passed in. Skipping this step.")
```
kaibolay, line 68: nit: different icon for "skip"?  🚫? ⏩? ⬜? 🗨️ ?

kaibolay (approved):
Customer ❤️!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: Good call! Went with ⏩

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -82,6 +89,9 @@ def self.run(params)
         if upload_status_response.binary_download_uri
           UI.message("🔗 Download the release binary (link expires in 1 hour): #{upload_status_response.binary_download_uri}")
         end
+
+        Actions.lane_context[:FIREBASE_APP_DISTRO_RELEASE] = release
+        release
```
Rezisn, line 94: love your work here, was thinking to do similar things
what do you think if we return `upload_status_response` here instead of just `release` ?
so something like `testing_url` and other things are accessible

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -71,16 +71,17 @@ def update_release_notes(release_name, release_notes)
...
             }
           }
-          connection.patch(update_release_notes_url(release_name), payload.to_json) do |request|
+          response = connection.patch(update_release_notes_url(release_name), payload.to_json) do |request|
             request.headers[AUTHORIZATION] = "Bearer " + @auth_token
             request.headers[CONTENT_TYPE] = APPLICATION_JSON
             request.headers[CLIENT_VERSION] = client_version_header_value
           end
+          UI.success("✅ Posted release notes.")
+          response.body
```
lfkellogg, line 80: 1. This method still supports nil or empty release notes, but it returns nil in that case. Since we're now always going to pass in notes I think we should stop handling that case.
2. We should update the documentation for this method to specify the new return value, and any other changes.

lfkellogg (approved):
Love this! Just one suggestion.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: 1. Good call! Forgot to remove that code. Done.
2. Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: Thanks for the comment @Rezisn! 

+1 to what @kaibolay said. All the useful information is part of the release, so decided not to return the entire `upload_status_response`.

PR #307, opened 2023-05-03
Prefer Google auth library and generated API client for list releases action

lfkellogg:
Use Google Auth library for application default credentials and use generated API client instead of hand rolled client in the list releases action.


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_get_latest_release.rb
```diff
@@ -12,24 +12,54 @@ class FirebaseAppDistributionGetLatestReleaseAction < Action
...
-          latest_release = releases[0]
+          # latest_release = append_json_style_fields(response.releases[0].to_h)
+          latest_release = map_release_hash(releases[0])
           UI.success("✅ Latest release fetched successfully. Returning release and setting Actions.lane_context[SharedValues::FIREBASE_APP_DISTRO_LATEST_RELEASE].")
         end
         Actions.lane_context[SharedValues::FIREBASE_APP_DISTRO_LATEST_RELEASE] = latest_release
         return latest_release
       end
 
+      def self.map_release_hash(release)
```
kaibolay, line 48: is this for backwards compatibility, because other fastlane plugins might rely on camelCase instead of snake_case?

Or is this temporary until you update the rest of the code?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -23,43 +25,45 @@ module FirebaseAppDistributionAuthClient
...
+        nil
+      end
+
       def refresh_token_from_firebase_tools
         config_path = format_config_path
         if File.exist?(config_path)
           begin
             firebase_tools_tokens = JSON.parse(File.read(config_path))['tokens']
-            if firebase_tools_tokens.nil?
-              UI.user_error!(ErrorMessage::EMPTY_TOKENS_FIELD)
```
kaibolay, line 60: why is this removed?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_get_latest_release.rb
lfkellogg: For backwards compatibility

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: I might move this to a separate PR. I realized that this isn't actually an exceptional case. If they log out of the Firebase CLI (at least the version I'm using), that file will still exist it just won't have a `tokens` field. In that case we just want to return `nil` and continue.

tagboola (approved):
👏🏿👏🏿👏🏿

Are you going to iteratively make the changes i.e. submit this one fix first and keep going or include all the changes in one big change?

lfkellogg:
> 👏🏿👏🏿👏🏿
> 
> Are you going to iteratively make the changes i.e. submit this one fix first and keep going or include all the changes in one big change?

Either. If we want to collaborate on this then I think splitting it up makes sense. The only downside of that is that we'd probably want to temporarily support both paths in the auth client. In this PR I am breaking the old path.

Issue #305, opened 2023-04-04
use official, auto-generated Google API client


kaibolay:
This plugin should use the official, auto-generated Google API client: [google-apis-firebaseappdistribution_v1](https://github.com/googleapis/google-api-ruby-client/tree/main/generated/google-apis-firebaseappdistribution_v1) ([RubyGems](https://rubygems.org/gems/google-apis-firebaseappdistribution_v1)) instead of the hand-written [firebase_app_distribution_api_client.rb](https://github.com/fastlane/fastlane-plugin-firebase_app_distribution/blob/master/lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb).

lfkellogg:
This was partially rolled back due to slow uploads: https://github.com/firebase/fastlane-plugin-firebase_app_distribution/pull/336

PR #309, opened 2023-05-23
Add functionality for creation of tester groups

garry-jeromson:
### Use case

Firebase App Distribution groups (see https://firebase.google.com/docs/reference/app-distribution/rest/v1/projects.groups/) are a useful resource in distributing releases; they allow indicating the use case that certain testers may have in a way that is clearer than a list of ungrouped testers.

### Implementation

Functionality for creating and deleting groups, and adding/removing testers to/from groups, has been added to the API client. The existing `add_testers` action now takes an additional, optional `group_alias` parameter to add testers to an existing group.

Feedback most welcome - this functionality would be really useful for us, and hopefully for others too.


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_tester_groups_action.rb
```diff
@@ -0,0 +1,105 @@
...
+          UI.user_error!("Invalid JSON file content. #{e.message}")
+        end
+
+        groups = tester_group_data["groups"]
+        groups.each do |group|
+          group_alias = group['alias']
+          group_display_name = group["displayName"]
+          UI.message("⏳ Creating tester group '#{group_alias}' testers in project #{params[:project_number]}...")
+
+          fad_api_client.add_tester_group(params[:project_number], group_alias, group_display_name)
```
tagboola, line 47: Hmm I wonder if we shouldn't conflate the action of adding testers to groups with creating groups. Maybe we should make this a simple create group action, and then modify the existing `firebase_app_distribution_add_testers` action to support adding testers to groups. @lfkellogg @kaibolay what do y'all think?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_tester_groups_action.rb
```diff
@@ -0,0 +1,105 @@
...
+      end
+
+      def self.available_options
+        [
+          FastlaneCore::ConfigItem.new(key: :project_number,
+                                       env_name: "FIREBASEAPPDISTRO_PROJECT_NUMBER",
+                                       description: "Your Firebase project number. You can find the project number in the Firebase console, on the General Settings page",
+                                       type: Integer,
+                                       optional: false),
+          FastlaneCore::ConfigItem.new(key: :json_file,
```
tagboola, line 79: Prefer just `file` to be consistent with the rest of the actions.

tagboola (commented):
Thanks for doing this! I left some preliminary high level comments to kick off the review.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
kaibolay, line 1: Instead of adding those hand-crafted consider #305 (with some initial implementation in #307), that is use the auto-generated client for the four API calls you're adding

kaibolay (commented):
Thanks for this PR! It's great to see community contributions.

I realize that my comments suggest some fundamental changes. Please let me know if you think they make sense.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola: I actually think we shouldn't conflate #307 with this change so I'd be fine with just using the hand crafted client endpoints for now 🙂

kaibolay: You're both right: Let's go with the hand-crafted code that's already written as part of this PR and move over to the auto-generated code as part of #307 when that's ready.

In the (probably unlikely) case that #307 is merged first, we should use the auto-generated code here, though.

lib/fastlane/plugin/firebase_app_distribution/helper/tester_groups_json_schema.rb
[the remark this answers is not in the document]
kaibolay: Yes, exactly.

kaibolay (approved):
Looks great. Thanks!

I assume the changes adding an optional `group_alias` parameter to `add_tester` will be done in a follow-up.

kaibolay (approved):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_delete_group_action.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+
+        project_number = params[:project_number]
+        group_alias = params[:alias]
+
+        UI.message("⏳ Deleting tester group '#{group_alias}' in project #{project_number}...")
+
+        fad_api_client.delete_group(project_number, group_alias)
+
+        UI.success("✅ Group created successfully.")
```
tagboola, line 28: *Group deleted successfully

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -251,6 +251,96 @@ def remove_testers(project_number, emails)
...
+      #   project_number - Firebase project number
+      #   group_alias - Alias of the tester group
+      #   emails - An array of emails to be added to the group.
+      #            A maximum of 1000 testers can be added at a time, if creating missing testers is enabled.
+      #   create_missing_testers - If true, missing testers will be created and added to the group.
+      #
+      def add_testers_to_group(project_number, group_alias, emails, create_missing_testers = true)
+        payload = { emails: emails,
+                    createMissingTesters: create_missing_testers }
+        UI.message("Adding testers to group URL: #{add_testers_to_group_url(project_number, group_alias)}")
```
tagboola, line 292: This message is redundant since we already log in the action.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -251,6 +251,96 @@ def remove_testers(project_number, emails)
...
+        payload = { emails: emails }
+        connection.post(remove_testers_from_group_url(project_number, group_alias), payload.to_json) do |request|
+          request.headers[AUTHORIZATION] = "Bearer " + @auth_token
+          request.headers[CONTENT_TYPE] = APPLICATION_JSON
+          request.headers[CLIENT_VERSION] = client_version_header_value
+        end
+      rescue Faraday::BadRequestError
+        UI.user_error!(ErrorMessage::INVALID_EMAIL_ADDRESS)
+      rescue Faraday::ResourceNotFound
+        UI.user_error!(ErrorMessage::INVALID_PROJECT)
```
tagboola, line 323: This should print `ErrorMessage::INVALID_TESTER_GROUP` since the group is the primary resource of this call.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -251,6 +251,96 @@ def remove_testers(project_number, emails)
...
+      #   group_alias - Alias of the tester group
+      #
+      def delete_group(project_number, group_alias)
+        connection.delete(delete_tester_group_url(project_number, group_alias)) do |request|
+          request.headers[AUTHORIZATION] = "Bearer " + @auth_token
+          request.headers[CONTENT_TYPE] = APPLICATION_JSON
+          request.headers[CLIENT_VERSION] = client_version_header_value
+        end
+      rescue Faraday::ResourceNotFound
+        UI.user_error!(ErrorMessage::INVALID_PROJECT)
```
tagboola, line 341: This should print `ErrorMessage::INVALID_TESTER_GROUP` since the group is the primary resource of this call.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -13,6 +13,8 @@ module ErrorMessage
   INVALID_PROJECT = "App Distribution could not find your Firebase project. Make sure to onboard an app in your project by pressing the \"Get started\" button on the App Distribution page in the Firebase console: https://console.firebase.google.com/project/_/appdistribution."
   INVALID_PATH = "Could not read content from"
   INVALID_TESTERS = "Could not enable access for testers. Check that the tester emails are formatted correctly, the groups exist and you are using group aliases (not group names) for specifying groups."
+  INVALID_TESTER_GROUP = "App Distribution could not find your tester group. Make sure that it exists before trying to add testers, and that the group alias is specified correctly."
+  INVALID_TESTER_GROUP_NAME = "The tester name group should be 4-63 characters, and valid characters are /[a-z][0-9]-/."
```
tagboola, line 17: * The _tester group name_ should be...

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -631,11 +631,255 @@
...
+        [
+          200,
+          {}, # response headers
+          {} # response body
+        ]
+      end
+
+      result = api_client.create_group(project_number, group_alias, display_name)
+
+      expect(result.success?).to eq(true)
```
tagboola, line 669: Consider returning the body of the response i.e. the `group` hash, and verifying that in the test.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -631,11 +631,255 @@
...
+        [
+          409,
+          {}, # response headers
+          {} # response body
+        ]
+      end
+
+      result = api_client.create_group(project_number, group_alias, display_name)
+
+      expect(result).to eq(true)
```
tagboola, line 687: Consider returning the body of the response i.e. the `group` hash, and verifying that in the test.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -631,11 +631,255 @@
...
+    it 'is successful' do
+      stubs.delete("/v1/projects/#{project_number}/groups/#{group_alias}", headers) do |env|
+        [
+          200,
+          {}, # response headers
+          {} # response body
+        ]
+      end
+
+      result = api_client.delete_group(project_number, group_alias)
```
tagboola, line 742: Consider returning the body of the response, which would actually be null, and verifying that in the test instead of returning the response wrapper object.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -631,11 +631,255 @@
...
+      }
+      stubs.post("/v1/projects/#{project_number}/groups/#{group_alias}:batchJoin", payload.to_json, headers) do |env|
+        [
+          200,
+          {}, # response headers
+          {} # response body
+        ]
+      end
+
+      result = api_client.add_testers_to_group(project_number, group_alias, emails)
```
tagboola, line 786: Consider returning the body of the response, which would actually be null, and verifying that in the test instead of returning the response wrapper object.

spec/firebase_app_distribution_api_client_spec.rb
```diff
@@ -631,11 +631,255 @@
...
+      payload = { emails: emails }
+      stubs.post("/v1/projects/#{project_number}/groups/#{group_alias}:batchLeave", payload.to_json, headers) do |env|
+        [
+          200,
+          {}, # response headers
+          {} # response body
+        ]
+      end
+
+      result = api_client.remove_testers_from_group(project_number, group_alias, emails)
```
tagboola, line 850: Consider returning the body of the response, which would actually be null, and verifying that in the test instead of returning the response wrapper object.

lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
tagboola, line 1: Are you planning on adding functionality for adding testers to a group in a separate PR? If so consider moving the changes here to the API client to add `remove_testers_from_group` and `add_testers_from_group` methods into that PR to keep this one smaller and focused.

tagboola (changes requested):
This looks incredible! Thanks again for doing this. Left a few comments 🙂

garry-jeromson:
Thanks for the feedback fellas! I've added the modifications to the existing `add_testers` action, where we can now pass an optional `group_alias` argument, to this PR - it's a pretty small change, and I think it's useful to have the context of our discussion here rather than creating a separate PR. 😊

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_add_testers_action.rb
```diff
@@ -66,6 +71,11 @@ def self.available_options
                                       description: "Path to a file containing a comma separated list of tester emails to be created. A maximum of 1000 testers can be deleted at a time",
                                       optional: true,
                                       type: String),
+          FastlaneCore::ConfigItem.new(key: :group_alias,
+                                       env_name: "FIREBASEAPPDISTRO_ADD_TESTERS_GROUP_ALIAS",
+                                       description: "Alias for the group to add the specified testers to. The group must already exist. If not specified, testers will not be added to a group",
```
tagboola, line 76: *Alias _of_ the group

tagboola (approved):

kaibolay (approved):

kaibolay:
Thank you! I'll bump the version a create a new release.

PR #312, opened 2023-06-05
Bump version for added groups support.

kaibolay:


lib/fastlane/plugin/firebase_app_distribution/version.rb
```diff
@@ -1,5 +1,5 @@
 module Fastlane
   module FirebaseAppDistribution
-    VERSION = "0.5.0"
+    VERSION = "0.5.1"
```
rebehe, line 3: Not a bug fix, should it be a minor bump?

kaibolay: You're right. Done.

rebehe (approved):

PR #313, opened 2023-06-07
Support `group_alias` parameter for `remove_testers_action`.

kaibolay:
This removes the testers from only the specified group, but doesn't delete them from the project.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
@@ -26,11 +26,15 @@ module Fastlane
           UI.user_error!("A maximum of 1000 testers can be removed at a time.")
         end
 
-        UI.message("⏳ Removing #{emails.count} testers from project #{params[:project_number]}...")
-
-        count = fad_api_client.remove_testers(params[:project_number], emails)
-
-        UI.success("✅ #{count} tester(s) removed successfully.")
+        if blank?(params[:group_alias])
+          UI.message("⏳ Removing #{emails.count} testers from project #{params[:project_number]}...")
+          count = fad_api_client.remove_testers(params[:project_number], emails)
+          UI.success("✅ #{count} tester(s) removed successfully.")
+        else
+          UI.message("⏳ Removing #{emails.count} testers from group #{params[:group_alias]}...")
+          fad_api_client.remove_testers_from_group(params[:project_number], params[:group_alias], emails)
+          UI.success("✅ Tester(s) removed successfully.")
+        end
       end
 
       def self.description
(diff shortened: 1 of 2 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
```diff
@@ -26,9 +26,13 @@ def self.run(params)
...
         end
 
-        UI.message("⏳ Removing #{emails.count} testers from project #{params[:project_number]}...")
-
-        count = fad_api_client.remove_testers(params[:project_number], emails)
+        if blank?(params[:group_alias])
+          UI.message("⏳ Removing #{emails.count} testers from project #{params[:project_number]}...")
+          count = fad_api_client.remove_testers(params[:project_number], emails)
+        else
+          UI.message("⏳ Removing testers from group #{params[:group_alias]}...")
```
tagboola, line 33: Consider specifying the count of testers in the message like in the case of removing testers from a project.

lfkellogg (approved):
👏👏👏👏

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
```diff
@@ -26,9 +26,13 @@ def self.run(params)
...
+          fad_api_client.remove_testers_from_group(params[:project_number], params[:group_alias], emails)
+        end
 
         UI.success("✅ #{count} tester(s) removed successfully.")
```
tagboola, line 37: `count` won't be defined here when testers are removed from group. Actually, will `count` even be defined when removing testers from the project since it's declared in the scope of the if/else? I guess Ruby may allow you to get away with that 🤔

tagboola (changes requested):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
kaibolay: Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
kaibolay: Fixed by removing the message inside the `if`/`else`

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_remove_testers_action.rb
```diff
@@ -26,11 +26,15 @@ def self.run(params)
...
-
-        UI.success("✅ #{count} tester(s) removed successfully.")
+        if blank?(params[:group_alias])
+          UI.message("⏳ Removing #{emails.count} testers from project #{params[:project_number]}...")
+          count = fad_api_client.remove_testers(params[:project_number], emails)
+          UI.success("✅ #{count} tester(s) removed successfully.")
+        else
+          UI.message("⏳ Removing #{emails.count} testers from group #{params[:group_alias]}...")
+          fad_api_client.remove_testers_from_group(params[:project_number], params[:group_alias], emails)
+          UI.success("✅ tester(s) removed successfully.")
```
tagboola, line 36: Nit: Should `testers(s)` be capitalized?

tagboola, line 36: Can we include the count of testers in the success message with `emails.count` or is that number not necessarily accurate?

kaibolay: Yes.

kaibolay: It's not an error to try to remove a tester that wasn't a member of the group. So the number might be inaccurate. Since a number is logged just one line above I think it's fine to omit it here - especially since it can be inaccurate.

PR #314, opened 2023-07-18
Prefer generated API client for create and delete group actions

tagboola:


lib/fastlane/plugin/firebase_app_distribution/client/firebase_app_distribution_api_client.rb
```diff
@@ -330,23 +300,6 @@ def remove_testers_from_group(project_number, group_alias, emails)
...
-          request.headers[AUTHORIZATION] = "Bearer " + @auth_token
-          request.headers[CONTENT_TYPE] = APPLICATION_JSON
-          request.headers[CLIENT_VERSION] = client_version_header_value
-        end
-        response.body
-      rescue Faraday::ResourceNotFound
-        UI.user_error!(ErrorMessage::INVALID_TESTER_GROUP)
-      end
-
       # List releases
```
kaibolay, line 303: Is this still in use? Or should it have been deleted in the previous PR?

kaibolay (approved):

PR #317, opened 2023-07-19
Update upload action to use generated api client to get AAB info

tagboola:


spec/firebase_app_distribution_action_spec.rb
```diff
@@ -221,6 +221,7 @@
     before(:each) do
       allow(File).to receive(:exist?).and_call_original
       allow(action).to receive(:fetch_auth_token).and_return('fake-auth-token')
```
kaibolay, line 223: is this still needed?

kaibolay (approved):
please fix b0rken spec

spec/firebase_app_distribution_action_spec.rb
tagboola: Oh yeah I can get rid of that good call. Done!

PR #318, opened 2023-07-20
Update upload and update release note methods to use generated API client

tagboola:


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -66,29 +68,32 @@ def self.run(params)
...
         end
 
-        if upload_status_response.binary_download_uri
-          UI.message("🔗 Download the release binary (link expires in 1 hour): #{upload_status_response.binary_download_uri}")
+        if release.binary_download_uri
+          UI.message("🔗 Download the release binary (link expires in 1 hour): #{release.binary_download_uri}")
         end
 
-        Actions.lane_context[SharedValues::FIREBASE_APP_DISTRO_RELEASE] = release
+        Actions.lane_context[SharedValues::FIREBASE_APP_DISTRO_RELEASE] = deep_symbolize_keys(JSON.parse(release.to_json))
```
kaibolay, line 96: just curious: what does this do / why is this needed?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -200,6 +205,72 @@ def self.release_notes(params)
...
+
+      def self.upload_binary(app_name, binary_path, client, timeout)
+        options = Google::Apis::RequestOptions.new
+        options.max_elapsed_time = timeout
+        options.header = {
+          'Content-Type' => 'application/octet-stream',
+          'X-Goog-Upload-File-Name' => File.basename(binary_path),
+          'X-Goog-Upload-Protocol' => 'raw'
+        }
+        # For some reason calling the client.upload_medium returns nil when
```
kaibolay, line 246: Ugh. Is the return value the only problem, i.e. does the upload itself work?

Would be really nice to figure this out. I miss @jladieu! We also have other Ruby experts: Maybe @jakeouellette can help? Or @i2amsam?

Or we ask the Googlers who wrote the Ruby client: @bajajneha27 @dazuma @sqrrrl

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: The release was initially represented as a hash with keys that are camel cased [symbols](https://www.rubyguides.com/2018/02/ruby-symbols/) e.g.

```
{
  name: string,
  displayVersion: string,
}
```

The generated API client returns a release object. If you call `to_h` on the release it creates a hash, but the fields are snake case. To ensure it's camel case, I call the `to_json` method which creates a json string with camel case keys and then use `JSON.parse` to convert it back to a hash. While this correctly creates a hash that has camel case keys, the keys in the hash are _strings instead of symbols_.

```
{
  "name": string,
  "displayVersion": string,
}
```

So to ensure we are returning the same type of hash and not breaking folks (e.g. `release[:name]` won't return the name if the keys are actually strings not symbols), we need to _symbolize_ the keys. Does that make sense?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: How dare you question my ruby skills! 😛 

I don't think this is necessarily a ruby problem but more of a google api client issues. The upload actually does work, and you can see the correct response in the debug log, but something weird must be happening when the response is being deserialized, which is causing the response to come back as _nil_. We need the operation name from the response to poll the status of the upload so we're screwed if we don't get the information from the response.

FWIW I don't think this is that bad of a workaround. We would have to set these custom headers anyway since I _believe_ we only support raw uploads and need to include the file name to determine if it's an APK or an AAB.  This is still better than what we were doing in the custom API client.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
kaibolay: Yes. Thank!

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
kaibolay: Since the upload works, I agree that it's probably a bug in the API client code.  My idea was to debug the API client code to find out what's wrong/weird and get that fixed. It should not return `nil`.

Our server _should_ support all types of uploads, since it's using Scotty.

As for the custom headers, I looked at the [code](https://github.com/googleapis/google-api-ruby-client/blob/main/google-apis-core/lib/google/apis/core/upload.rb) and unfortunately `X-Goog-Upload-File-Name` is not set. I guess we could/should fix that. The code does set the `Content-Type` header based on the file extension, though. So we could update our server-side code to fall back to that if there's no filename.

I guess I'm fine with leaving a `TODO(kbolay): ...` for now.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -200,6 +205,72 @@ def self.release_notes(params)
         release_notes_param || Actions.lane_context[SharedValues::FL_CHANGELOG]
       end
 
+      def self.poll_release(client, operation, binary_type)
```
kaibolay, line 208: nit: better name, maybe `poll_operation` or `poll_upload_operation` or `poll_upload_release_operation`?

kaibolay (approved):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
tagboola: Done.

PR #325, opened 2023-07-25
Handle user errors when uploading releases

tagboola:


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -271,6 +271,39 @@ def self.release_version(release)
...
+          UI.crash!(err)
+        end
+      end
+
+      def self.distribute_release(client, release, request)
+        client.distribute_project_app_release(release.name, request)
+      rescue Google::Apis::Error => err
+        case err.status_code.to_i
+        when 400
+          UI.user_error!("#{ErrorMessage::INVALID_TESTERS}\nEmails: #{request.tester_emails} \nGroup Aliases: #{request.group_aliases}")
```
kaibolay, line 301: why include tester emails and group aliases in the error message?

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
```diff
@@ -15,7 +15,7 @@ module ErrorMessage
   INVALID_TESTERS = "Could not enable access for testers. Check that the tester emails are formatted correctly, the groups exist and you are using group aliases (not group names) for specifying groups."
   INVALID_TESTER_GROUP = "App Distribution could not find your tester group. Make sure that it exists before trying to add testers, and that the group alias is specified correctly."
   INVALID_TESTER_GROUP_NAME = "The tester group name should be 4-63 characters, and valid characters are /[a-z][0-9]-/."
-  INVALID_RELEASE_NOTES = "Failed to add release notes"
+  INVALID_RELEASE_NOTES = "Failed to update release notes"
```
kaibolay, line 18: There's no stand alone action/lane to update release notes, so they are "added" as part of uploading a new release. But I understand that the "add" in the old error message might not always be correct (since a duplicate upload might have happened - in which case the release notes are actually updated).

How about the neutral "Failed to set release notes"?

But that all nitpicking, so 🤷 if you disagree

kaibolay (approved):

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_error_message.rb
tagboola: Neutrality wins! Done.

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
kaibolay: Well, let's keep it. I somehow missed that this was already done before.

PR #336, opened 2023-08-03
Use old API client for uploads and bump to 0.7.2.pre.1

lfkellogg:
Reverts the switchover for uploads to the new client in https://github.com/firebase/fastlane-plugin-firebase_app_distribution/pull/318. This should address https://github.com/firebase/fastlane-plugin-firebase_app_distribution/issues/330.

Confirmed locally that it reduces upload times for a large binary from 120s down to 30s.

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
@@ -38,6 +38,9 @@ module Fastlane
         elsif !ENV["FIREBASE_TOKEN"].nil? && !ENV["FIREBASE_TOKEN"].empty?
           UI.message("🔐 Authenticating with FIREBASE_TOKEN environment variable")
           firebase_token(ENV["FIREBASE_TOKEN"], debug)
+        elsif !ENV["GOOGLE_APPLICATION_CREDENTIALS"].nil? && !ENV["GOOGLE_APPLICATION_CREDENTIALS"].empty?
+          UI.message("🔐 Authenticating with GOOGLE_APPLICATION_CREDENTIALS environment variable: #{ENV['GOOGLE_APPLICATION_CREDENTIALS']}")
+          service_account(ENV["GOOGLE_APPLICATION_CREDENTIALS"], debug)
         elsif !application_default_creds.nil?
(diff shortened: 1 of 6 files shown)


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -38,6 +38,9 @@ def get_authorization(google_service_path, firebase_cli_token, debug = false)
         elsif !ENV["FIREBASE_TOKEN"].nil? && !ENV["FIREBASE_TOKEN"].empty?
           UI.message("🔐 Authenticating with FIREBASE_TOKEN environment variable")
           firebase_token(ENV["FIREBASE_TOKEN"], debug)
+        elsif !ENV["GOOGLE_APPLICATION_CREDENTIALS"].nil? && !ENV["GOOGLE_APPLICATION_CREDENTIALS"].empty?
+          UI.message("🔐 Authenticating with GOOGLE_APPLICATION_CREDENTIALS environment variable: #{ENV['GOOGLE_APPLICATION_CREDENTIALS']}")
```
rebehe, line 42: If we're not supporting ADC right now, what's the point of this? Is this confusing?

spec/firebase_app_distribution_action_spec.rb
```diff
@@ -318,9 +318,9 @@ def stub_get_aab_info(integration_state = 'INTEGRATED')
 
         it 'crashes if it exceeds polling threshold' do
           stub_const('Fastlane::Actions::FirebaseAppDistributionAction::MAX_POLLING_RETRIES', 0)
-          allow_any_instance_of(Google::Apis::FirebaseappdistributionV1::FirebaseAppDistributionService)
-            .to receive(:http)
-            .and_return({ name: 'operation-name' }.to_json)
```
rebehe, line 323: Wouldn't we still need this mock for the non-upload calls?

rebehe (approved):

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: Good point. I've updated to make it a little clearer.

PR #347, opened 2023-12-30
Use generated API clients

kaibolay:
Use auto-generated API client (#305) for getting tester UDIDs and upload (again, reverting c313e1c6c45cfe5dd564be5a3b2bdbc847bcadca).

### Upload performance using the generated client has been fixed
 
The previous code passed the binary to be uploaded as a String object (instead of an IO object) as the request body. This significantly slowed upload performance, since the SSL socket sends 16kB at a time and resizes the String buffer with each iteration. This caused CPU to skyrocket, since it requires allocating a new buffer and copying existing data into that string. The garbage collector had to work hard to keep up.

To eliminate this unnecessary String resizing, we now simply pass the `File` object (which is an `IO` object) as the `body`, This enables `OpenSSL::Buffering` to read and write a full 16K buffer. No resizing (allocating, copying, garbage collecting) necessary.

### Check/use cached Firebase CLI credential before Application Default Credentials.
    
On Cloudtop, the metadata server returns a default service account, so authentication with Application Default Credentials is attempted. But that account doesn't have any scopes, so authentication fails. The cached Firebase CLI credentials are not used in that case.
    
By changing the order, authenticating with the cached Firebase CLI credentials works on Cloudtops.


lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
```diff
@@ -56,6 +52,12 @@ def get_authorization(google_service_path, firebase_cli_token, debug = false)
 
       private
 
+      def application_default_creds
+        Google::Auth.get_application_default([SCOPE])
```
lfkellogg, line 56: Just to be extra careful, did you confirm that setting `GOOGLE_APPLICATION_CREDENTIALS` to a "service account key file" (JSON) as described in the docs still works? https://firebase.google.com/docs/app-distribution/ios/distribute-fastlane#service-acc-fastlane. I can help with this if you want.

Side note: after this ships we should be able to update the docs to specify that we support ADC in general.

lfkellogg (approved):
Amazing!

.tool-versions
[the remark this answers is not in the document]
kaibolay: I'm using `adsf` (because `rbenv` fails to compile Ruby on my Linux workstation). This indeed caused the update of this file. I think you're right: This shouldn't affect anyone not using `asdf`.

lib/fastlane/plugin/firebase_app_distribution/helper/firebase_app_distribution_auth_client.rb
lfkellogg: Confirmed locally that this still works 👍

PR #350, opened 2024-01-11
Miscellaneous cleanup.

kaibolay:
- consistently end messages with period (except in cases where a sentence ends with a colon followed by a value
- sleep before polling for upload status (because the server needs some time for asynchronous processing after upload)
- reorder/refactor some code


lfkellogg (approved):
Nice!

Issue #376, opened 2024-11-22
Firebase deployment failed: Couldn't find binary.


anton-demydov-zoral:
### Issue Description
Deployment fails with the error Firebase deployment failed: Couldn't find binary.

##### Configuration
```
  lane :firebase_deploy_android do
    UI.message "Deploying Android build to Firebase"
    ENV['GOOGLE_APPLICATION_CREDENTIALS'] = File.expand_path('./firebase-key.json')

    apk_path = Actions.lane_context[SharedValues::GRADLE_APK_OUTPUT_PATH]
    UI.message "APK Path: #{apk_path}"
    UI.message "Firebase Android App ID: #{ENV['FIREBASE_ANDROID_APP_ID']}"

    unless File.exist?(apk_path)
      UI.error "APK file not found at #{apk_path}"
      UI.user_error!("Failed to locate the APK file for Firebase deployment.")
    end

    begin
      firebase_app_distribution(
        android_artifact_path: apk_path,
        android_artifact_type: "APK",
        app: ENV['FIREBASE_ANDROID_APP_ID'],
        service_credentials_file: './firebase-key.json', # Optional: Use service account JSON
        debug: true
      )
      UI.success "Successfully deployed to Firebase"
    rescue => e
      UI.error "Firebase deployment failed: #{e.message}"
      UI.message "Continuing build despite Firebase deployment failure"
    end
  end
```

##### Complete verbose output
```
[09:12:20]: ----------------------------------------
[09:12:20]: --- Step: Verifying fastlane version ---
[09:12:20]: ----------------------------------------
[09:12:20]: Your fastlane version 2.225.0 matches the minimum requirement of 2.198.0  ✅
...
| fastlane-plugin-firebase_app_distribution | 0.9.1   | firebase_app_distribution_add_testers, firebase_app_distribution_get_udids, firebase_app_distribution_create_group, firebase_app_distribution_delete_group, firebase_app_distribution, firebase_app_distribution_get_latest_release, firebase_app_distribution_remove_testers     
....
     FIREBASE_ANDROID_APP_ID: ***
    FIREBASE_IOS_APP_ID: ***
    FIREBASE_CLI_TOKEN: ***
    GOOGLE_APPLICATION_CREDENTIALS: $HOME/firebase-key.json

....

 [09:37:12]: Deploying Android build to Firebase
[...]

PR #377, opened 2024-11-29
Log platform if binary not found

kaibolay:


kaibolay:
Maybe this helps debugging #376

tonybaroneee (approved):

PR #382, opened 2024-12-12
Support `test_case_ids` and `test_case_ids_file`

kaibolay:
Bump version to 0.10.0

### DIFF
--- a/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
+++ b/lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
@@ -57,7 +57,7 @@ module Fastlane
 
         binary_type = binary_type_from_path(binary_path)
         UI.message("📡 Uploading the #{binary_type}.")
-        operation = upload_binary(app_name, binary_path, binary_type, client, timeout)
+        operation = upload_binary(client, app_name, binary_path, binary_type, timeout)
         UI.message("🕵️ Validating upload…")
         release = poll_upload_release_operation(client, operation, binary_type)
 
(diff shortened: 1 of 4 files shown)


lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -401,37 +405,64 @@ def self.test_release(alpha_client, release, test_devices, username = nil, passw
...
-            device_executions: device_executions
+            test_case: test_case_name
           )
-        alpha_client.create_project_app_release_test(release.name, release_test)
+        alpha_client.create_project_app_release_test(release_name, release_test)
       rescue Google::Apis::Error => err
-        UI.crash!(err)
+        case err.status_code.to_i
+        when 404
+          UI.user_error!("Test Case #{test_case_name} not found")
```
lfkellogg, line 431: Can we guarantee that any 404 means the test case wasn't found, and not the release/app/project?

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
```diff
@@ -401,37 +405,64 @@ def self.test_release(alpha_client, release, test_devices, username = nil, passw
...
+            if release_test.device_executions.all? { |e| e.state == 'PASSED' }
+              true
             else
-              UI.test_failure!("Unsupported automated test state for #{device_to_s(de.device)}: #{de.state}.")
+              release_test.device_executions.each do |de|
+                case de.state
+                when 'PASSED', 'IN_PROGRESS'
+                  next
+                when 'FAILED'
+                  UI.test_failure!("Automated test failed for #{device_to_s(de.device)}: #{de.failed_reason}.")
```
lfkellogg, line 452: Nice! Didn't know about `test_failure!`

lfkellogg (approved):

lib/fastlane/plugin/firebase_app_distribution/actions/firebase_app_distribution_action.rb
kaibolay: No. I made the [same comment](https://github.com/FirebasePrivate/firebase-android-buildtools/pull/545#discussion_r1880499200) when Tunde implemented this for Gradle.

But since it's extremely unlikely that the 404 is because the release doesn't exist (after all it was just created/updated due to the upload) and extremely likely that this is caused my a missing test case, I was convinced to add this.

lfkellogg: Fair!

